PR #1, opened 2021-05-19
Document DSL format

dcreager:
These patches are just focusing on documenting the DSL format, and having something approaching a specification of its execution.

### DIFF
--- a/src/reference/mod.rs
+++ b/src/reference/mod.rs
@@ -325,3 +325,29 @@
 //!   }
 //! }
 //! ```
+//!
+//! # Functions
+//!
+//! The process executing a graph DSL file can provide **_functions_** that can be called from
+//! within graph DSL stanzas.
+//!
+//! Function calls use a Lisp-like syntax, where the name of the function being called is _inside_
+//! of the parentheses.  The parameters to a function call are arbitrary expressions.  For
+//! instance, if the executing process provides a function named `+`, you could call it as:
+//!
+//! ``` tsg
+//! (identifier) @id
+//! {
+//!    let x = 4
+//!    attr @id.node "nine" = (+ x 5)
+//! }
+//! ```
+//!
+//! Note that it's the process executing the graph DSL file that decides which functions are
+//! available.  We do define a [standard library][], and most of the time those are the functions
+//! that are available, but you should double-check the documentation of whatever graph DSL tool
+//! you're using to make sure.
+//!
+//! [standard library]: functions/index.html
+
+pub mod functions;
--- a/src/reference/mod.rs
+++ b/src/reference/mod.rs
@@ -1,0 +1,210 @@
+// -*- coding: utf-8 -*-
+// ------------------------------------------------------------------------------------------------
+// Copyright © 2021, tree-sitter authors.
+// Licensed under either of Apache License, Version 2.0, or MIT license, at your option.
+// Please see the LICENSE-APACHE or LICENSE-MIT files in this distribution for license details.
+// ------------------------------------------------------------------------------------------------
+
+//! This section defines the graph DSL implemented by this library.
+//!
+//! [tree-sitter]: https://docs.rs/tree-sitter/
+//! [tree-sitter-python]: https://docs.rs/tree-sitter-python/
+//!
+//! # Overview
+//!
+//! You can use [tree-sitter][] to parse the content of source code into a _concrete syntax tree_.
+//! For instance, using the [tree-sitter-python][] grammar, you can parse Python source code:
+//!
+//! ``` python
+//! # test.py
+//! from one.two import d, e.c
+//! import three
+//! print(d, e.c)
+//! print three.f
+//! ```
+//!
+//! ``` text
+//! $ tree-sitter parse test.py
+//! (module [0, 0] - [4, 0]
+//!   (import_from_statement [0, 0] - [0, 26]
+//!     module_name: (dotted_name [0, 5] - [0, 12]
+//!       (identifier [0, 5] - [0, 8])
+//!       (identifier [0, 9] - [0, 12]))
+//!     name: (dotted_name [0, 20] - [0, 21]
+//!       (identifier [0, 20] - [0, 21]))
+//!     name: (dotted_name [0, 23] - [0, 26]
+//!       (identifier [0, 23] - [0, 24])
+//!       (identifier [0, 25] - [0, 26])))
+//!   (import_statement [1, 0] - [1, 12]
+//!     name: (dotted_name [1, 7] - [1, 12]
+//!       (identifier [1, 7] - [1, 12])))
+//!   (expression_statement [2, 0] - [2, 13]
+//!     (call [2, 0] - [2, 13]
+//!       function: (identifier [2, 0] - [2, 5])
+//!       arguments: (argument_list [2, 5] - [2, 13]
+//!         (identifier [2, 6] - [2, 7])
+//!         (attribute [2, 9] - [2, 12]
+//!           object: (identifier [2, 9] - [2, 10])
+//!           attribute: (identifier [2, 11] - [2, 12])))))
+//!   (print_statement [3, 0] - [3, 13]
+//!     argument: (attribute [3, 6] - [3, 13]
+//!       object: (identifier [3, 6] - [3, 11])
+//!       attribute: (identifier [3, 12] - [3, 13]))))
+//! ```
+//!
+//! There are many interesting use cases where you want to use this parsed syntax tree to create
+//! some _other_ graph-like structure.  This library lets you do that, using a declarative DSL to
+//! identify patterns in the parsed syntax tree, along with rules for which nodes and edges to
+//! create for the syntax nodes that match those patterns.
+//!
+//! # Limitations
+//!
+//! You can create any graph structure, as long as each node "belongs to" or "is associated with"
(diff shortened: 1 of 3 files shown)


dcreager:
The best way to view the rendered documentation would be to clone this locally and run

```
$ cargo doc --no-deps --open
```

src/reference/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+//! # Syntax nodes
+//!
+//! Syntax nodes are identified by tree-sitter query captures (`@name`).  For instance, in our
+//! example stanza, whose query is `(identifier) @id`, `@id` would refer to the `identifier` node
+//! that the stanza matched against.
+//!
+//! # Graph nodes
+//!
+//! Each node in the output graph is identified by a syntax node and a list of **_tag names_**.
+//! Graph nodes are automatically created when they are referenced.  Graph node expressions have
```
rewinfrey, line 140: > Graph nodes are automatically created when they are referenced.

If I'm following correctly, this means that a graph node is automatically created when a _syntax node_ is referenced? I think the word _they_ in this context makes this a bit ambiguous.

src/reference/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+//! ```
+//!
+//! Multiple stanzas can add attributes to the same graph node or edge.  Any attributes defined in
+//! those stanzas are **_combined_** into a single set of attributes by taking the union of their
+//! contents.  If both stanzas define an attribute with a particular name, then the corresponding
+//! values must either be:
+//!
+//!   - both sets, in which case the values are unioned together
+//!   - anything else, in which case the values must be equal
+//!
```
rewinfrey, line 208: What's the syntax like for a set in this DSL? I wonder if this would be helpful to add as an example in the `# Expressions` section above (along with an example showing an ordered list of values)?

src/reference/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+//!     values carry over from stanza to stanza.  Scoped variables are referenced by using a syntax
+//!     node expression (typically a query capture) and a variable name, separated by a
+//!     double-colon: `@node::variable`.
+//!
+//! Local and scoped variables are created using `var` or `let` statements.  A `let` statement
+//! creates an **_immutable variable_**, whose value cannot be changed.  A `var` statement creates
+//! a **_mutable variable_**.  You use a `set` statement to change the value of a mutable variable.
+//!
+//! (All global variables are immutable, and cannot be created by any graph DSL statement; they are
+//! only provided by the external process the executes the graph DSL file.  If you need to create
```
rewinfrey, line 234:
```suggestion
//! only provided by the external process that executes the graph DSL file.  If you need to create
```

src/reference/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+//!       ; module's name comes from the last directory component.
+//!
+//!       ; Expose the graph node that we created for that component as a
+//!       ; scoped variable that later stanzas can see.
+//!       let @mod::root = current
+//!     }
+//!
+//!     "([^/]+)\\.py$"
+//!     {
+//!       ; This arm will any other trailing module name.  Note that
```
rewinfrey, line 317:
```suggestion
//!       ; This arm will match any other trailing module name.  Note that
```

src/reference/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+//! groups in the regular expression.
+//!
+//! For example, if `filepath` is a global variable containing the path of a Python source file,
+//! you could use the following `scan` statement to construct graph nodes for the name of the
+//! module defined in the file:
+//!
+//! ``` tsg
+//! (module) @mod
+//! {
+//!   var current = @mod.root
```
rewinfrey, line 289: Does `root` have special significance in this DSL, or is this just a tag?

src/reference/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+//!   }
+//! }
+//! ```
+//!
+//! # Functions
+//!
+//! The process executing a graph DSL file can provide **_functions_** that can be called from
+//! within graph DSL stanzas.
+//!
+//! Functions calls use a Lisp-like syntax, where the name of the function being called is _inside_
```
rewinfrey, line 334:
```suggestion
//! Function calls use a Lisp-like syntax, where the name of the function being called is _inside_
```

src/reference/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+//! of the parentheses.  The parameters to a function call are arbitrary expressions.  For
+//! instance, if the executing process provides a function named `+`, you could call it as:
+//!
+//! ``` tsg
+//! (identifier) @id
+//! {
+//!    let x = 4
+//!    attr @id.node "nine" = (+ x 5)
```
rewinfrey, line 342: Does this imply that each syntax node / tag pair has distinct attributes?

For example:

```shell
(identiifer) @id
{
  let x = 4
  attr @id.node1 "nine" = (+ x 5)
  attr @id.node2 "ten" = (+ x 6)
}
```

Then when referencing the `@id` graph node, there would be two tags, and each tag would have one attribute associated with it?

rewinfrey (commented):
👍  This was very helpful and fun to read. Thanks for the excellent write up. I left a few clarification questions and a couple small tweaks to wording.

rewinfrey:
I'm not sure if I'm holding this incorrectly, but when I try `$ cargo doc --no-deps --open`, I'm not able to find or see the documentation. This screenshot below shows what my local env is producing:

image

src/reference/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+//!   - `[graph node].[tag name]`
+//!
+//! For instance, all of the following are valid graph node expressions in our example stanza:
+//!
+//!   - `@id.node`
+//!   - `@id.node.another`
+//!
+//! For both expressions, the graph node's syntax node would be whichever node was captured by
+//! `@id` in the stanza's query.  They would have `node` and `node.another`, respectively, as their
+//! tag names.
```
tausbn, line 153: I read this as meaning that the two `.`s in `@id.node.another` are different, the first delineating the syntax node `@id` from its tag, and the second just being a part of the tag `node.another`.

Would it make sense to perhaps exclude `.` as an allowed character in tags, in case you want to assign a different meaning to `@id.node.another` in the future? 

Also, there should probably be a specification of what kinds of characters are permitted in tag names.

src/reference/mod.rs
dcreager: `@id` isn't a graph node, it's a syntax node.  `@id.node1` and `@id.node2` are two distinct graph nodes, both "attached" to the `@id` syntax node, and each graph node has one attribute.

The whole description of syntax / graph nodes could use some rewording.

src/reference/mod.rs
dcreager: Just a tag name like any other

src/reference/mod.rs
dcreager: Good suggestions, thanks!

src/reference/mod.rs
dcreager: Added a description of this and lists

dcreager:
I've taken a stab at rewording the text about syntax nodes and graph nodes, please take a look and let me know if it's clearer.

I've also made some changes in the direction of making everything more explicit than implicit:

- Attribute values are now _never_ combined implicitly, even if they're sets.  We're not using that right now, and if we ever need to, we can add a new statement that lets you modify the value of an attribute however you want.

- Graph nodes and edges are no longer created implicitly.  You must explicitly use a `node` or `edge` statement to create them.  You can't create a node or edge more than once, and you can't add an attribute to a node or edge that doesn't exist.

Thoughts?

rewinfrey (approved):
These updates are ✨  thank you, this dispels several ambiguities I had about the graph DSL.

src/reference/mod.rs
BekaValentine: Each graph node is generated in virtue of being referenced, but also each syntax node is a graph node.

src/reference/mod.rs
BekaValentine: Technically `@id` is also a graph node, since the list of tags can be empty. It's not uncommon to do stuff like `edge @foo -> @foo.bar`, at least insofar as anything can be common or uncommon currently.

src/reference/mod.rs
BekaValentine: The current implementation of the DSL is such that tags are lists of strings, not just single strings, so that if you have a tagged node `(n, ts)`, the syntax for that is `"@" ++ concat (map ("."++) ts)`. So you get `.` as the separator between each of the tag parts.

The general idea behind this is that the tags are intended to kind of.. associated nodes with other nodes in a convenient way. For instance, you might have a value associated with a node `@node.value` and that value might have a scope that it's relative to `@node.value.scope`, or so forth.

It can be removed but it would be a breaking change wrt current code.

BekaValentine:
I think the proposed syntax changes are good, I've always found the implicit creation of nodes to be tricky, especially from the perspective of what the semantics of these things are. It adds some noise to do it this way, but I think that's ok, because it gives us more ability to catch errors statically. A not uncommon problem is, if you typo a node tag, it's fine, you implicitly create that node and it's no problem. Having to declare the node upfront gives a single place where nodes are formed, and references to them are *merely* references, which can be scope checked, which is good.

PR #2, opened 2021-05-21
Graph nodes not implicitly linked to syntax nodes

dcreager:
This is an alternative version of #1 that _doesn't_ use the "tag name" mechanism to implicitly attach each graph node to exactly one syntax node.  Instead you can just create graph nodes whenever you want, and use the existing "scoped variables" mechanism to attach them to syntax nodes whenever you want.

### DIFF
--- a/src/reference/mod.rs
+++ b/src/reference/mod.rs
@@ -178,7 +178,7 @@
 //!
 //!   - **_Scoped_** variables are "attached to" syntax nodes.  Their values carry over from stanza
 //!     to stanza.  Scoped variables are referenced by using a syntax node expression (typically a
-//!     query capture) and a variable name, separated by a double-colon: `@node::variable`.
+//!     query capture) and a variable name, separated by a period: `@node.variable`.
 //!
 //! Local and scoped variables are created using `var` or `let` statements.  A `let` statement
 //! creates an **_immutable variable_**, whose value cannot be changed.  A `var` statement creates
@@ -203,7 +203,7 @@
 //!   var mutable_variable = "first value"
 //!   set mutable_variable = "second value"
 //!
-//!   var @id::kind = "id"
+//!   var @id.kind = "id"
 //! }
 //! ```
 //!
@@ -225,7 +225,7 @@
 //! (identifier) @id
 //! {
 //!    let x = 4
-//!    let @id::nine = (+ x 5)
+//!    let @id.nine = (+ x 5)
 //! }
 //! ```
 //!
@@ -262,13 +262,13 @@
 //! ``` tsg
 //! (identifier) @id
 //! {
-//!   let @id::node = (node)
+//!   let @id.node = (node)
 //! }
 //!
 //! (dotted_name (identifier) @dotted_element)
 //! {
 //!   ; We will learn more about the attr statement below
-//!   attr (@dotted_element::node) kind = "dotted"
+//!   attr (@dotted_element.node) kind = "dotted"
 //! }
 //! ```
 //!
@@ -288,9 +288,9 @@
 //! ``` tsg
 //! (import_statement name: (_) @name)
 //! {
-//!   let @name::source = (node)
-//!   let @name::sink = (node)
-//!   edge @name::source -> @name::sink
+//!   let @name.source = (node)
+//!   let @name.sink = (node)
+//!   edge @name.source -> @name.sink
 //! }
 //! ```
 //!
@@ -308,10 +308,10 @@
 //! ``` tsg
 //! (import_statement name: (_) @name)
 //! {
-//!   let @name::source = (node)
-//!   let @name::sink = (node)
-//!   attr (@name::sink) kind = "module"
-//!   attr (@name::source -> @name::sink) precedence = 10
+//!   let @name.source = (node)
+//!   let @name.sink = (node)
+//!   attr (@name.sink) kind = "module"
+//!   attr (@name.source -> @name.sink) precedence = 10
 //! }
 //! ```
 //!
@@ -369,7 +369,7 @@
 //!
 //!       ; Expose the graph node that we created for that component as a
 //!       ; scoped variable that later stanzas can see.
-//!       let @mod::root = current_node
+//!       let @mod.root = current_node
 //!     }
 //!
 //!     "([^/]+)\\.py$"
@@ -381,7 +381,7 @@
 //!       set new_node = (node)
 //!       attr (new_node) name = $1
 //!       edge current_node -> new_node
-//!       let @mod::root = new_node
+//!       let @mod.root = new_node
 //!     }
 //!   }
 //! }
--- a/src/reference/mod.rs
+++ b/src/reference/mod.rs
@@ -65,19 +65,6 @@
 //! use "syntax node" and "graph node" in this documentation to make it clear which kind of node we
 //! mean.
 //!
-//! # Limitations
-//!
-//! You can create any graph structure, as long as each graph node "belongs to" or "is associated
-//! with" exactly one syntax node in the concrete syntax tree.  There are no limitations on how you
-//! use edges to connect the graph nodes: you are not limited to creating a tree, and in
-//! particular, you are not limited to creating a tree that "lines" up with the parsed syntax tree.
-//!
-//! You can annotate each graph node and edge with an arbitrary set of attributes.  As we will see
-//! below, many stanzas in the graph DSL can contribute to the set of attributes for a particular
-//! graph node or edge.  Each stanza must define attributes in a "consistent" way — you cannot have
-//! multiple stanzas provide conflicting values for a particular attribute for a particular graph
-//! node or edge.
-//!
 //! # High-level structure
 //!
 //! A graph DSL file consists of one or more **_stanzas_**.  Each stanza starts with a tree-sitter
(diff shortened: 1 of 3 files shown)


dcreager:
@BekaValentine @rewinfrey @tausbn Interested in your thoughts on this

src/reference/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+//! to the new node.  You will almost always assign that result to a variable so that you can then
+//! do interesting things with the node:
+//!
+//! [`node`]: functions/index.html#node
+//!
+//! ``` tsg
+//! (identifier) @id
+//! {
+//!   let new_node = (node)
+//! }
```
rewinfrey, line 256: As @tausbn noted, I agree that a shorthand `node new_node` syntax that desugars to `let new_node = (node)` would be great for consistency and ergonomics.

src/reference/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+//! ``` tsg
+//! (identifier) @id
+//! {
+//!   let @id::node = (node)
+//! }
+//!
+//! (dotted_name (identifier) @dotted_element)
+//! {
+//!   ; We will learn more about the attr statement below
+//!   attr (@dotted_element::node) kind = "dotted"
```
rewinfrey, line 271: Clarifying question: are the parenthesis around `(@dotted_element::node)` required?

src/reference/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+//! Graph nodes and edges have an associated set of **_attributes_**.  Each attribute has a name
+//! (which is an identifier), and a value.
+//!
+//! You add attributes to a graph node or edge using an `attr` statement:
+//!
+//! ``` tsg
+//! (import_statement name: (_) @name)
+//! {
+//!   let @name::source = (node)
+//!   let @name::sink = (node)
```
rewinfrey, line 312: Small nit: it looks like these graph nodes were declared in the previous stanza. Are we wanting this stanza to be independent of the previous one?

rewinfrey (commented):
I like it 👍 , I think dropping the `tag` portion of this DSL is a nice simplification. Copying over some of the chat from Slack, I also vote for @tausbn's suggestion for the shorthand syntax for declaring graph nodes.

dcreager:
@maxbrunsfeld I'd be interested in your thoughts on this too!  We're working on pulling the graph DSL out into a version that can be used for other use cases than stack graphs.  This PR is trying to put together some reference documentation before moving any of the implementation over.  While we're at it I'm trying to clean up some of the under-specified parts.

#1 was an original stab at it that retains the `[capture].[tag name].[tag name]` syntax for implicitly creating graph nodes.  This PR is an alternative spec that changes the language a bit more, but in a way that I think is more understandable and usable.

src/reference/mod.rs
dcreager: Yes, as currently written, but for aesthetic reasons not to make the parsing easier.

I had put that in more because of the edge attribute statement, so that it's clearer which part of the statement is the edge reference and which part is the list of attributes:

```
attr (@id::source -> @id::sink) precedence = 10
```

and then thought I should keep the parens for node attributes, too, for consistency.

src/reference/mod.rs
dcreager: Yeah in this case I had tried to make the stanza independent, but I hear your point that it's not entirely clear whether that's true of all the stanzas in the doc, or not.

maxbrunsfeld:
This design seems really elegant! I like the idea of using the same `::` (scoped variable) syntax for all relationships between syntax nodes and graph nodes, instead of having the separate `.` syntax. Maybe if you're going to do that, you should repurpose the `.` operator to replace `::`. Just because `.` is shorter and easier to type 😁 .

src/reference/mod.rs
```diff
@@ -0,0 +1,404 @@
...
+//!
+//! Identifiers start with either an ASCII letter or underscore, and all remaining characters are
+//! ASCII letters, numbers, underscores, or hyphens.  (More precisely, they satisfy the regular
+//! expression `/[a-zA-Z_][a-zA-Z0-9_-]*/`.)  Identifiers are used as the names of
+//! [attributes](#attributes), [functions](#functions), and [variables](#variables).
+//!
+//! To execute a graph DSL file against a concrete syntax tree, we execute each stanza in the graph
+//! DSL file _in order_.  For each stanza, we identify each place where the concrete syntax tree
+//! matches the query pattern.  For each of these places, we end up with a different set of syntax
+//! nodes assigned to the query pattern's captures.  We execute the block of statements for each of
```
maxbrunsfeld, line 90: This description of the order in which the stanzas execute might not match the behavior of the system we implemented. Maybe that's intentional though?

In case it's helpful, the thing we originally built (based on the single-pass query execution) processed the matches in the order that they were encountered in the syntax tree during a depth-first traversal. Specifically, a match would be processed as soon as the *last* node in its pattern was found.

src/reference/mod.rs
tausbn: I think @rewinfrey highlights a salient point here. If `(node)` creates a new node, what does the following code do?
```
(identifier) @id
{
  let @id::node = (node)
}

(identifier) @id
{
  let @id::node = (node)
}
```
- Is it an error? And will this be flagged at runtime -- when the second assignment is attempted -- or at "compile time"?
- Does it silently reuse the same node for both `let`s? (Thus preserving the notion of immutability that `let` implies?)
- Does it create two distinct `node` nodes for each `identifier`? (I mean, almost certainly not. That seems like a terrible idea.)

If the `let @id::node = (node)` declaration can appear in multiple different stanzas, then this creates the possibility that a typo creates a completely unrelated node (which may not be apparent until things start breaking).

If the `let @id::node = (node)` declaration can appear in at most one stanza (for a given scoped variable), then this creates a somewhat awkward situation where one stanza becomes responsible for ensuring these nodes are created (and possibly creates some edges and attributes between these), and all other stanzas do not. (Which means if you want to get rid of that stanza, you have to make sure to push the declarations into another appropriate stanza lest everything breaks.)

Perhaps what is needed is a way to say "these scoped variables are available on these `tree-sitter` nodes", so above we would say `identifier` can have a `node` scoped variable. Then if we write `node @id::node`, we're actually saying "oh, and please make sure to emit a graph node for this scoped variable", but crucially it'll be the same graph node across all accesses. And typos should be prevented, since we can check each scoped variable access against the set of scoped variables defined on its corresponding `tree-sitter` node.

(This would also make it easy to see exactly what scoped variables are defined for a given `tree-sitter` node. Currently you would have to find all captures that mention that node and look in the associated stanza to see what scoped variables are set on it.)

I'm not entirely sure how this would interact with mutability, though.

src/reference/mod.rs
tausbn: I think it would be better if the order of execution (between stanzas) was left unspecified. Saying it's in the order in which they appear in the file means you have to keep the entire context in your head when thinking about this (which feels very imperative). This does, however, raise the question of _when_ (or rather _where_) to define the graph nodes that are emitted.

Having the insides of each stanza be executed in order seems fine, however, and indeed required for some of the more complex stanzas (like the one that handles Python modules).

src/reference/mod.rs
dcreager: >     * Is it an error? And will this be flagged at runtime -- when the second assignment is attempted -- or at "compile time"?

This is the current semantics, and it doesn't matter whether it's a new node reference that you're assigning to the variable. In general, you can only assign to an immutable `let` variable once, across all of the stanzas of the file.  So this would be an error, as well:

```
(identifier) @id
{
  let @id::count = 9;
}

(identifier) @id
{
  ; This is invalid
  let @id::count = 17;
}
```

There's a similar story with mutable variables: you can only declare them with `var` once, and you can't `set` them unless you've already declared them.  Although the `var` declaration and `set` reassignment don't have to be in the same stanza:

```
; This is valid

(identifier) @id
{
  var @id::count = 9;
}

(identifier) @id
{
  set @id::count = 17;
}
```

src/reference/mod.rs
dcreager: > This description of the order in which the stanzas execute might not match the behavior of the system we implemented. Maybe that's intentional though?

Yep, it's intentionally different:

> Saying it's in the order in which they appear in the file means you have to keep the entire context in your head when thinking about this (which feels very imperative). This does, however, raise the question of _when_ (or rather _where_) to define the graph nodes that are emitted.

It actually was originally unspecified, and as Max points out, in the current implementation we throw every stanza's query pattern into a single `TSQuery` instance, which means that the matches for each stanza can be interleaved together.  We've actually found that to be _harder_ to reason about, especially for larger graph DSL files.  I suggested moving to this specified order, where you're _guaranteed_ that earlier stanzas execute _fully_ before later stanzas, because it enables thinking about the stanzas locally, since the execution order depends only on what's in the graph DSL file, and not on the dynamic content of the input file you're processing.  And it's how we were structuring the graph DSL files anyway.

> In case it's helpful, the thing we originally built (based on the single-pass query execution) processed the matches in the order that they were encountered in the syntax tree during a depth-first traversal. Specifically, a match would be processed as soon as the _last_ node in its pattern was found.

I might need to reword how matches _within_ a stanza are ordered based on this.  The proposed new implementation is that each stanza would get its own `TSQuery` instance, and within a stanza we'll get matches however they come out of the query engine.

(@maxbrunsfeld, as an aside, this was the _actual_ prompt for this change! We wanted separate `TSQuery` instances anyway to try to make "match limit exceeded" errors less likely, before realizing that meant we could specify a particular execution order for stanzas.)

dcreager:
I've incorporated the immediately actionable feedback:

- scoped variable syntax now uses `.` instead of `::`
- added a `node` statement that's syntactic sugar for the common `let foo = (node)` use case
- called out the different between attributes and scoped variables

The other bit of discussion was around how I've explicitly specified the order of execution between stanzas. I feel strongly about keeping this part, since it was one of the major areas that has made it hard to think about our stack graph DSL files.

I'm going to go ahead and merge this to get a baseline spec into `main`.  If there are any follow-on changes that we need to make, we can do that in separate issues or PRs.

PR #4, opened 2021-05-27
Add graph DSL parser

dcreager:
This is ported over from the internal version, modified to include the language changes introduced recently.

Co-authored-by: Max Brunsfeld <maxbrunsfeld@gmail.com>
Co-authored-by: Patrick Thomson <patrickt@github.com>
Co-authored-by: Rick Winfrey <rewinfrey@github.com>

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -1,0 +1,364 @@
+// -*- coding: utf-8 -*-
+// ------------------------------------------------------------------------------------------------
+// Copyright © 2021, tree-sitter authors.
+// Licensed under either of Apache License, Version 2.0, or MIT license, at your option.
+// Please see the LICENSE-APACHE or LICENSE-MIT files in this distribution for license details.
+// ------------------------------------------------------------------------------------------------
+
+//! Defines the AST structure of a graph DSL file
+
+use regex::Regex;
+use tree_sitter::Language;
+use tree_sitter::Query;
+
+use crate::Identifier;
+
+/// A graph DSL file
+#[derive(Debug)]
+pub struct File {
+    pub language: Language,
+    /// The list of stanzas in the file
+    pub stanzas: Vec<Stanza>,
+}
+
+impl File {
+    pub fn new(language: Language) -> File {
+        File {
+            language,
+            stanzas: Vec::new(),
+        }
+    }
+}
+
+/// One stanza within a file
+#[derive(Debug)]
+pub struct Stanza {
+    /// The tree-sitter query for this stanza
+    pub query: Query,
+    /// The list of statements in the stanza
+    pub statements: Vec<Statement>,
+}
+
+/// A statement that can appear in a graph DSL stanza
+#[derive(Debug, Eq, PartialEq)]
+pub enum Statement {
+    // Variables
+    DeclareImmutable(DeclareImmutable),
+    DeclareMutable(DeclareMutable),
+    Assign(Assign),
+    // Graph nodes
+    CreateGraphNode(CreateGraphNode),
+    AddGraphNodeAttribute(AddGraphNodeAttribute),
+    // Edges
+    CreateEdge(CreateEdge),
+    AddEdgeAttribute(AddEdgeAttribute),
+    // Regular expression
+    Scan(Scan),
+    // Debugging
+    Print(Print),
+}
+
+/// An `attr` statement that adds an attribute to an edge
+#[derive(Debug, Eq, PartialEq)]
+pub struct AddEdgeAttribute {
+    pub source: Expression,
+    pub sink: Expression,
+    pub attributes: Vec<Attribute>,
+}
+
+impl From<AddEdgeAttribute> for Statement {
+    fn from(statement: AddEdgeAttribute) -> Statement {
+        Statement::AddEdgeAttribute(statement)
+    }
+}
+
+/// An `attr` statement that adds an attribute to a graph node
+#[derive(Debug, Eq, PartialEq)]
+pub struct AddGraphNodeAttribute {
+    pub node: Expression,
+    pub attributes: Vec<Attribute>,
+}
+
+impl From<AddGraphNodeAttribute> for Statement {
+    fn from(statement: AddGraphNodeAttribute) -> Statement {
+        Statement::AddGraphNodeAttribute(statement)
+    }
+}
+
+/// A `set` statement that updates the value of a mutable variable
+#[derive(Debug, Eq, PartialEq)]
+pub struct Assign {
+    pub variable: Variable,
+    pub value: Expression,
+}
+
+impl From<Assign> for Statement {
+    fn from(statement: Assign) -> Statement {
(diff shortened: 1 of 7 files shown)


src/ast.rs
```diff
@@ -0,0 +1,364 @@
...
+#[derive(Debug, Eq, PartialEq)]
+pub enum Variable {
+    Scoped(ScopedVariable),
+    Unscoped(UnscopedVariable),
+}
+
+/// A reference to a scoped variable
+#[derive(Debug, Eq, PartialEq)]
+pub struct ScopedVariable {
+    pub scope: Box<Expression>,
```
robrix, line 209: Do we need arbitrary expressions as the scope of an identifier? e.g. string/integer scopes seem surprising?

src/ast.rs
```diff
@@ -0,0 +1,364 @@
...
+        UnscopedVariable { name }.into()
+    }
+}
+
+/// An expression that can appear in a graph DSL file
+#[derive(Debug, Eq, PartialEq)]
+pub enum Expression {
+    // Literals
+    FalseLiteral,
+    NullLiteral,
```
robrix, line 242: What are the semantics of null?

src/lib.rs
```diff
@@ -27,3 +27,46 @@
 
 #[cfg(doc)]
```
robrix, line 28: Should this file be renamed to something involving interning/contexts/whatever?

src/parser.rs
```diff
@@ -0,0 +1,589 @@
...
+    }
+}
+
+impl Display for Location {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "({}, {})", self.row + 1, self.column + 1)
+    }
+}
+
+struct Parser<'a> {
```
robrix, line 80: Is the rest of this generated code, or should someone (probably not me 😛) review it exhaustively?

tests/it/main.rs
```diff
@@ -4,3 +4,5 @@
 // Licensed under either of Apache License, Version 2.0, or MIT license, at your option.
 // Please see the LICENSE-APACHE or LICENSE-MIT files in this distribution for license details.
 // ------------------------------------------------------------------------------------------------
+
+mod parser;
```
robrix, line 8: Ok 👍🏻

tests/it/parser.rs
```diff
@@ -0,0 +1,225 @@
...
+// Copyright © 2021, tree-sitter authors.
+// Licensed under either of Apache License, Version 2.0, or MIT license, at your option.
+// Please see the LICENSE-APACHE or LICENSE-MIT files in this distribution for license details.
+// ------------------------------------------------------------------------------------------------
+
+use tree_sitter_graph::ast::*;
+use tree_sitter_graph::Context;
+
+#[test]
+fn can_parse_blocks() {
```
robrix, line 12: Would love to see smaller, expression- or statement-focused tests if possible, but otherwise this seems reasonable (and I am glad to see any tests at all 👍🏻).

robrix (approved):
Looks good! A couple of questions but otherwise 👍🏻

src/ast.rs
```diff
@@ -0,0 +1,364 @@
...
+    // Graph nodes
+    CreateGraphNode(CreateGraphNode),
+    AddGraphNodeAttribute(AddGraphNodeAttribute),
+    // Edges
+    CreateEdge(CreateEdge),
+    AddEdgeAttribute(AddEdgeAttribute),
+    // Regular expression
+    Scan(Scan),
+    // Debugging
+    Print(Print),
```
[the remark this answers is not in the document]
dcreager: The DSL as a whole is fairly imperative at this point.  We had some discussions in the PR about the reference documentation about this, e.g. https://github.com/tree-sitter/tree-sitter-graph/pull/2#discussion_r637396616 and https://github.com/tree-sitter/tree-sitter-graph/pull/2#issuecomment-847864935

I'm open to suggestions, though I think they might be best served in follow-on PRs that touch both the documentation and the implementation here.

src/ast.rs
```diff
@@ -0,0 +1,364 @@
...
+        Statement::DeclareImmutable(statement)
+    }
+}
+
+/// A `var` statement that declares a new mutable variable
+#[derive(Debug, Eq, PartialEq)]
+pub struct DeclareMutable {
+    pub variable: Variable,
+    pub value: Expression,
+}
```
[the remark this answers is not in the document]
dcreager: We use them within stanzas in the [`scan` statement example](https://docs.rs/tree-sitter-graph/0.0.1/tree_sitter_graph/reference/index.html#regular-expressions) to walk through the sequence of matches from a regexp.

We're also using in the stack graph rules in a couple of places to build up a linked list of graph nodes, and to annotate a parent with the tail of the list.

src/ast.rs
```diff
@@ -0,0 +1,364 @@
...
+        self.regex.as_str() == other.regex.as_str() && self.statements == other.statements
+    }
+}
+
+/// A reference to a variable
+#[derive(Debug, Eq, PartialEq)]
+pub enum Variable {
+    Scoped(ScopedVariable),
+    Unscoped(UnscopedVariable),
+}
```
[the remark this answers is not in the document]
dcreager: It means that the variable is attached to a syntax node: https://docs.rs/tree-sitter-graph/0.0.1/tree_sitter_graph/reference/index.html#variables

src/ast.rs
dcreager: It has to be an expression that evaluates to a syntax node reference, but it's checked at runtime: https://github.com/tree-sitter/tree-sitter-graph/blob/3550c82d382ebd329dc54eee10dcfcef29439c2e/src/execution.rs#L538-L541

src/ast.rs
dcreager: There don't really have semantics in the graph DSL itself, but they're there because you might want them as attributes that you attach to node and edges.  (I'm basically aiming for the JSON data model there, since that's the output format we'll use in an as-yet-unwritten PR)

src/lib.rs
dcreager: I can move the context stuff into its own separate file, but the root file of a Rust crate has to be called `lib.rs`

src/parser.rs
dcreager: It's not generated, but it is copy-pasted over from the internal repo where it was reviewed previously...  :smile:

tests/it/parser.rs
dcreager: Yeah the wrinkle is that as written, this test case can only use the crate's public API, and I purposefully didn't expose parsing functions for all of the internal non-terminals in the DSL parser.  I could maybe write helper functions in here that wrap up an expression in a minimal "surrounding file" or something like that, but it wasn't obviously worth the effort compared to just making sure there's instances of all interesting DSL patterns that we might want to parse.

PR #6, opened 2021-06-11
Add graph data model

dcreager:
This defines the graph structure of nodes, edges, and attributes, which is built up while executing a graph DSL file.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -1,0 +1,343 @@
+// -*- coding: utf-8 -*-
+// ------------------------------------------------------------------------------------------------
+// Copyright © 2021, tree-sitter authors.
+// Licensed under either of Apache License, Version 2.0, or MIT license, at your option.
+// Please see the LICENSE-APACHE or LICENSE-MIT files in this distribution for license details.
+// ------------------------------------------------------------------------------------------------
+
+//! Defines data types for the graphs produced by the graph DSL
+
+use std::collections::BTreeSet;
+use std::collections::HashMap;
+use std::fmt::Display;
+use std::ops::Index;
+use std::ops::IndexMut;
+
+use smallvec::SmallVec;
+use tree_sitter::Node;
+
+use crate::Context;
+use crate::Identifier;
+
+/// A graph produced by executing a graph DSL file.  Graphs include a lifetime parameter to ensure
+/// that they don't outlive the tree-sitter syntax tree that they are generated from.
+#[derive(Default)]
+pub struct Graph<'tree> {
+    syntax_nodes: HashMap<SyntaxNodeRef, Node<'tree>>,
+    graph_nodes: Vec<GraphNode>,
+}
+
+type SyntaxNodeID = u32;
+type GraphNodeID = u32;
+
+impl<'tree> Graph<'tree> {
+    /// Creates a new, empty graph.
+    pub fn new() -> Graph<'tree> {
+        Graph::default()
+    }
+
+    /// Adds a syntax node to the graph, returning a graph DSL reference to it.
+    ///
+    /// The graph won't contain _every_ syntax node in the parsed syntax tree; it will only contain
+    /// those nodes that are referenced at some point during the execution of the graph DSL file.
+    pub fn add_syntax_node(&mut self, node: Node<'tree>) -> SyntaxNodeRef {
+        let index = SyntaxNodeRef(node.id() as SyntaxNodeID);
+        self.syntax_nodes.insert(index, node);
+        index
+    }
+
+    /// Adds a new graph node to the graph, returning a graph DSL reference to it.
+    pub fn add_graph_node(&mut self) -> GraphNodeRef {
+        let graph_node = GraphNode::new();
+        let index = self.graph_nodes.len() as GraphNodeID;
+        self.graph_nodes.push(graph_node);
+        GraphNodeRef(index)
+    }
+
+    /// Displays the contents of this graph.
+    pub fn display_with<'a>(&'a self, ctx: &'a Context) -> impl Display + 'a {
+        struct DisplayGraph<'a, 'tree>(&'a Graph<'tree>, &'a Context);
+
+        impl<'a, 'tree> Display for DisplayGraph<'a, 'tree> {
+            fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+                let graph = self.0;
+                let ctx = self.1;
+                for (node_index, node) in graph.graph_nodes.iter().enumerate() {
+                    write!(
+                        f,
+                        "node {}\n{}",
+                        node_index,
+                        node.attributes.display_with(ctx, graph)
+                    )?;
+                    for (sink, edge) in &node.outgoing_edges {
+                        write!(
+                            f,
+                            "edge {} -> {}\n{}",
+                            node_index,
+                            *sink,
+                            edge.attributes.display_with(ctx, graph)
+                        )?;
+                    }
+                }
+                Ok(())
+            }
+        }
+
+        DisplayGraph(self, ctx)
+    }
+}
+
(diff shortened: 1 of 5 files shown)


dcreager:
This depends on #4, and shouldn't be reviewed until that is merged

src/graph.rs
```diff
@@ -0,0 +1,343 @@
...
+        GraphNode {
+            outgoing_edges: SmallVec::new(),
+            attributes: Attributes::new(),
+        }
+    }
+
+    /// Adds an edge to this node.  There can be at most one edge connecting any two graph nodes;
+    /// the result indicates whether the edge is new (`Ok`) or already existed (`Err`).  In either
+    /// case, you also get a mutable reference to the [`Edge`][] instance for the edge.
+    pub fn add_edge(&mut self, sink: GraphNodeRef) -> Result<&mut Edge, &mut Edge> {
```
robrix, line 128: Instead of failing when the edge already exists, could we union the attributes of the extant and new `Edge` and return successfully?

src/graph.rs
```diff
@@ -0,0 +1,343 @@
...
+        match self
+            .outgoing_edges
+            .binary_search_by_key(&sink, |(sink, _)| *sink)
+        {
+            Ok(index) => Err(&mut self.outgoing_edges[index].1),
+            Err(index) => {
+                self.outgoing_edges.insert(index, (sink, Edge::new()));
+                Ok(&mut self.outgoing_edges[index].1)
+            }
+        }
```
robrix, line 139: Could we use `get_edge` or `get_edge_mut` (below) to reduce the duplication in the search process?

src/graph.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    /// Returns a mutable reference to an outgoing edge from this node, if it exists.
+    pub fn get_edge_mut(&mut self, sink: GraphNodeRef) -> Option<&mut Edge> {
+        let sink = sink.0;
+        self.outgoing_edges
+            .binary_search_by_key(&sink, |(sink, _)| *sink)
+            .ok()
+            .map(move |index| &mut self.outgoing_edges[index].1)
+    }
```
robrix, line 158: Likewise here, is there a convenient means of collapsing the two definitions to avoid duplicating the binary search? If we can’t define one in terms of the other, could we internalize the iteration by passing in a closure (ideally non-escaping so it can be stack-allocated &c.) or something like that?

src/graph.rs
```diff
@@ -0,0 +1,343 @@
...
+    /// Adds an attribute to this attribute set.  If there was already an attribute with the same
+    /// name, replaces its value and returns `Err`.
+    pub fn add<V: Into<Value>>(&mut self, name: Identifier, value: V) -> Result<(), ()> {
+        match self.values.binary_search_by_key(&name, |(name, _)| *name) {
+            Ok(_) => Err(()),
+            Err(index) => {
+                self.values.insert(index, (name, value.into()));
+                Ok(())
+            }
+        }
```
robrix, line 197: 1. The docs appear to be incorrect; this seems to return `Err` without replacing the value.
2. As above with the graph itself, this seems hard to use correctly. It doesn’t look like we’d have a well-defined union on values in general, but maybe better semantics in this case would involve replacing the value (as the docs suggest) or unioning the values when well-defined (e.g. `Set`/`Set` values could be unioned)?
3. Again as above, can we factor the binary search so we don’t have to be vigilant about updating every other position as the code changes?
4. I believe `Result<(), ()>` is isomorphic to a boolean; if we want to continue reporting collisions, could we return the result as a boolean instead?

tests/it/graph.rs
```diff
@@ -0,0 +1,68 @@
...
+// Please see the LICENSE-APACHE or LICENSE-MIT files in this distribution for license details.
+// ------------------------------------------------------------------------------------------------
+
+use indoc::indoc;
+use tree_sitter::Parser;
+use tree_sitter_graph::graph::Graph;
+use tree_sitter_graph::Context;
+
+#[test]
+fn can_display_graph() {
```
robrix, line 14: This is a good thing to test 👍🏻

robrix (approved):
I’ve changed the base branch to point at the parent PR so that reviewing this made sense, and then, well, reviewed it. I’ve left a bunch of questions about factoring and returns which might conflict with idiomatic Rust, so do with it what you will. Good work 👍🏻

src/graph.rs
```diff
@@ -0,0 +1,343 @@
...
+        struct DisplayGraph<'a, 'tree>(&'a Graph<'tree>, &'a Context);
+
+        impl<'a, 'tree> Display for DisplayGraph<'a, 'tree> {
+            fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+                let graph = self.0;
+                let ctx = self.1;
+                for (node_index, node) in graph.graph_nodes.iter().enumerate() {
+                    write!(
+                        f,
+                        "node {}\n{}",
```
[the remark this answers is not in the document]
dcreager: Yep yep, that's definitely on the list — we have dot output on the internal repo, just have to think about how best to handle it in this more generic case.  Probably some well-defined attributes that you can assign to each node / edge to affect what styling info ends up in the dot output.

src/graph.rs
dcreager: This has come up a couple of times now.  If we went with a more Datalog-inspired DSL, where the execution order of stanzas isn't important, then this kind of unioning of content would be an important part of that.  But as written so far, the language is purposefully more imperative in nature, and so I wanted to minimize the amount of work that's done transparently for you.  I'd suggest that in that version of the language, unioning a set of attributes should be a function that we provide that you can call yourself when (and only when) you want that behavior.

src/graph.rs
dcreager: There's no way in Rust to generalize across `mut`/not `mut`, unfortunately.  For me, these method are small enough and near enough to each other that the code duplication isn't something that I worry about.  `binary_search_by_key` is a common enough part of the stdlib that I think it's a readable idiom.

src/graph.rs
dcreager: We can't as written because of how the `Err` branches are handled differently.  I could pull the `binary_search_by_key` invocation out into a separate method...let me see how that looks.  I mention below that this amount of copy/paste doesn't bother me, but that one particular refactoring might be the right balance between our views.

src/graph.rs
dcreager: >     1. The docs appear to be incorrect; this seems to return `Err` without replacing the value.

Good catch!  I'll fix that.



>     4\. I believe `Result<(), ()>` is isomorphic to a boolean; if we want to continue reporting collisions, could we return the result as a boolean instead?

I went with `Result<(), ()>` because then the caller will have some useful methods on the result that will make handling the two cases more ergonomic.  In particular `Result` works with the `?` operator, but a `bool` doesn't.

src/graph.rs
```diff
@@ -0,0 +1,343 @@
...
+                    Value::Null => write!(f, "#null"),
+                    Value::Boolean(value) => {
+                        if *value {
+                            write!(f, "#true")
+                        } else {
+                            write!(f, "#false")
+                        }
+                    }
+                    Value::Integer(value) => write!(f, "{}", value),
+                    Value::String(value) => write!(f, "{}", value),
```
tausbn, line 267:
```suggestion
                    Value::String(value) => write!(f, "\"{}\"", value.escape_default()),
```
This makes the output less confusing when the string contains a newline, and also disambiguates e.g. the integer `5` from the string `"5"`.

dcreager: It does: https://github.com/tree-sitter/tree-sitter-graph/blob/5461b8bf988caecde48650153ea652c49ba0faf8/tests/it/graph.rs#L69-L71

PR #7, opened 2021-06-11
Add graph DSL interpreter

dcreager:
This patch adds the execution rules for all of the graph DSL constructs, including the standard library of functions.

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -100,6 +100,11 @@ impl Functions {
         functions.add(ctx.add_identifier("start-column"), stdlib::StartColumn);
         functions.add(ctx.add_identifier("end-row"), stdlib::EndRow);
         functions.add(ctx.add_identifier("end-column"), stdlib::EndColumn);
+        functions.add(ctx.add_identifier("node-type"), stdlib::NodeType);
+        functions.add(
+            ctx.add_identifier("named-child-count"),
+            stdlib::NamedChildCount,
+        );
         functions
     }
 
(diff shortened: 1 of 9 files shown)


dcreager:
This depends on #6, and shouldn't be reviewed until that is merged

src/functions.rs
```diff
@@ -0,0 +1,308 @@
...
+    impl Function for StartRow {
+        fn call(
+            &mut self,
+            graph: &mut Graph,
+            _source: &str,
+            parameters: &mut dyn Parameters,
+        ) -> Result<Value, ExecutionError> {
+            let node = parameters.param()?.into_syntax_node(graph)?;
+            parameters.finish()?;
+            Ok(Value::String(node.start_position().row.to_string()))
```
tausbn, line 253:
```suggestion
            Ok(Value::Integer(node.start_position().row as u32))
```

src/functions.rs
```diff
@@ -0,0 +1,308 @@
...
+    impl Function for EndRow {
+        fn call(
+            &mut self,
+            graph: &mut Graph,
+            _source: &str,
+            parameters: &mut dyn Parameters,
+        ) -> Result<Value, ExecutionError> {
+            let node = parameters.param()?.into_syntax_node(graph)?;
+            parameters.finish()?;
+            Ok(Value::String(node.end_position().row.to_string()))
```
tausbn, line 288:
```suggestion
            Ok(Value::Integer(node.end_position().row as u32))
```

src/functions.rs
```diff
@@ -0,0 +1,308 @@
...
+    impl Function for StartColumn {
+        fn call(
+            &mut self,
+            graph: &mut Graph,
+            _source: &str,
+            parameters: &mut dyn Parameters,
+        ) -> Result<Value, ExecutionError> {
+            let node = parameters.param()?.into_syntax_node(graph)?;
+            parameters.finish()?;
+            Ok(Value::String(node.start_position().column.to_string()))
```
tausbn, line 271:
```suggestion
            Ok(Value::Integer(node.start_position().column as u32))
```

src/functions.rs
```diff
@@ -0,0 +1,308 @@
...
+    impl Function for EndColumn {
+        fn call(
+            &mut self,
+            graph: &mut Graph,
+            _source: &str,
+            parameters: &mut dyn Parameters,
+        ) -> Result<Value, ExecutionError> {
+            let node = parameters.param()?.into_syntax_node(graph)?;
+            parameters.finish()?;
+            Ok(Value::String(node.end_position().column.to_string()))
```
tausbn, line 305:
```suggestion
            Ok(Value::Integer(node.end_position().column as u32))
```

tausbn (commented):
My implementations of the position-accessing functions returned strings instead of integers, which was less than ideal. I have added suggestions that fix this.

src/functions.rs
```diff
@@ -0,0 +1,348 @@
...
+        functions.add(ctx.add_identifier("child-index"), stdlib::ChildIndex);
+        functions.add(ctx.add_identifier("node"), stdlib::Node);
+        functions.add(ctx.add_identifier("plus"), stdlib::Plus);
+        functions.add(ctx.add_identifier("replace"), stdlib::Replace);
+        functions.add(ctx.add_identifier("source-text"), stdlib::SourceText);
+        functions.add(ctx.add_identifier("start-row"), stdlib::StartRow);
+        functions.add(ctx.add_identifier("start-column"), stdlib::StartColumn);
+        functions.add(ctx.add_identifier("end-row"), stdlib::EndRow);
+        functions.add(ctx.add_identifier("end-column"), stdlib::EndColumn);
+        functions.add(ctx.add_identifier("node-type"), stdlib::NodeType);
```
dcreager, line 103: @tausbn I brought in your changes from the `a-few-minor-fixes` branch.  I took the liberty of renaming the `kind` function to `node-type` — it's only called `kind` in the Rust bindings because `type` is a reserved word.  The underlying concept in tree-sitter itself is "[node type](https://tree-sitter.github.io/tree-sitter/using-parsers#syntax-nodes)".

tausbn: That's perfect, thank you! (Though actually I ended up not needing the `kind`/`node-type` function. I guess it's useful to keep around regardless, though.)

tausbn (approved):
Looks good to me! :+1:

PR #8, opened 2021-06-11
Add initial `tree-sitter-graph` CLI

dcreager:
This patch implements a `tree-sitter-graph` CLI program that parses a source file and executes a graph DSL program against it.  It prints out a textual representation of the resulting graph.  In future PRs, we'll look at defining a machine-readable JSON format for the graph.

The CLI program uses tree-sitter's existing logic for determining the language of a source file, and loading its grammar.  It's your responsibility to make sure that the graph DSL program is compatible with that language.


tausbn (approved):
Looks good! :+1:

PR #10, opened 2021-10-15
Improve error messages

hendrikvanantwerpen:
This PR introduces changes that improve error messages thrown during execution by including variable and attribute names, and node values when these are available.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -8,9 +8,12 @@
 //! Defines the AST structure of a graph DSL file
 
 use regex::Regex;
+use std::fmt;
 use tree_sitter::Language;
 use tree_sitter::Query;
 
+use crate::Context;
+use crate::DisplayWithContext;
 use crate::Identifier;
 
 /// A graph DSL file
@@ -58,6 +61,22 @@ pub enum Statement {
     Print(Print),
 }
 
+impl DisplayWithContext for Statement {
+    fn fmt(&self, f: &mut fmt::Formatter, ctx: &Context) -> fmt::Result {
+        match self {
+            Statement::DeclareImmutable(stmt) => stmt.fmt(f, ctx),
+            Statement::DeclareMutable(stmt) => stmt.fmt(f, ctx),
+            Statement::Assign(stmt) => stmt.fmt(f, ctx),
+            Statement::CreateGraphNode(stmt) => stmt.fmt(f, ctx),
+            Statement::AddGraphNodeAttribute(stmt) => stmt.fmt(f, ctx),
+            Statement::CreateEdge(stmt) => stmt.fmt(f, ctx),
+            Statement::AddEdgeAttribute(stmt) => stmt.fmt(f, ctx),
+            Statement::Scan(stmt) => stmt.fmt(f, ctx),
+            Statement::Print(stmt) => stmt.fmt(f, ctx),
+        }
+    }
+}
+
 /// An `attr` statement that adds an attribute to an edge
 #[derive(Debug, Eq, PartialEq)]
 pub struct AddEdgeAttribute {
@@ -69,6 +88,21 @@ pub struct AddEdgeAttribute {
 impl From<AddEdgeAttribute> for Statement {
     fn from(statement: AddEdgeAttribute) -> Statement {
         Statement::AddEdgeAttribute(statement)
+    }
+}
+
+impl DisplayWithContext for AddEdgeAttribute {
+    fn fmt(&self, f: &mut fmt::Formatter, ctx: &Context) -> fmt::Result {
+        write!(
+            f,
+            "attr ({} -> {})",
+            self.source.display_with(ctx),
+            self.sink.display_with(ctx),
+        )?;
+        for attr in &self.attributes {
+            write!(f, " {}", attr.display_with(ctx))?;
+        }
+        write!(f, "")
     }
 }
 
@@ -85,6 +119,16 @@ impl From<AddGraphNodeAttribute> for Statement {
     }
 }
 
+impl DisplayWithContext for AddGraphNodeAttribute {
+    fn fmt(&self, f: &mut fmt::Formatter, ctx: &Context) -> fmt::Result {
+        write!(f, "attr ({})", self.node.display_with(ctx),)?;
+        for attr in &self.attributes {
+            write!(f, " {}", attr.display_with(ctx),)?;
+        }
+        write!(f, "")
+    }
+}
+
 /// A `set` statement that updates the value of a mutable variable
 #[derive(Debug, Eq, PartialEq)]
 pub struct Assign {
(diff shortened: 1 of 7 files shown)


src/graph.rs
```diff
@@ -354,6 +302,78 @@ impl From<String> for Value {
...
+            }
+            Value::Set(values) => {
+                write!(f, "{{")?;
+                for elem in values {
+                    write!(f, "{},", elem)?;
+                }
+                write!(f, "}}")
+            }
+            Value::SyntaxNode(value) => <dyn fmt::Display>::fmt(value, f),
+            Value::GraphNode(value) => <dyn fmt::Display>::fmt(value, f),
```
hendrikvanantwerpen, line 327: Compiler complained about "multiple applicable items in scope" when writing `value.fmt(f)` here. I couldn't figure out which items that were, but calling it like this fixed that. Alternatively, we could `write!("{}", value)`, but this would add another indirection.

src/graph.rs
```diff
@@ -354,6 +302,78 @@ impl From<String> for Value {
...
+        }
+    }
+}
+
+impl DisplayWithGraph for Value {
```
hendrikvanantwerpen, line 332: This trait implementation replaces the method that was implemented directly on `Value`.

src/execution.rs
```diff
@@ -48,6 +51,7 @@ impl File {
     /// provide the set of functions and global variables that are available during execution.
     pub fn execute<'tree>(
         &self,
+        ctx: &Context,
```
hendrikvanantwerpen, line 54: This new parameter is a breaking change for external callers.

src/graph.rs
```diff
@@ -354,6 +302,78 @@ impl From<String> for Value {
     }
 }
 
+impl fmt::Display for Value {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Value::Null => write!(f, "#null"),
+            Value::Boolean(value) => write!(f, "{}", value),
+            Value::Integer(value) => write!(f, "{}", value),
+            Value::String(value) => write!(f, "\"{}\"", value),
```
dcreager, line 311: If you use this, it will add the quotes automatically, and also escape non-printable characters in the string:

```suggestion
            Value::String(value) => write!(f, "{:?}", value),
```

src/graph.rs
```diff
@@ -354,6 +302,78 @@ impl From<String> for Value {
...
+                }
+                write!(f, "]")
+            }
+            Value::Set(values) => {
+                write!(f, "{{")?;
+                for elem in values {
+                    write!(f, "{},", elem)?;
+                }
+                write!(f, "}}")
+            }
```
dcreager, line 325: Can we retain the behavior from before, where commas were used as separators instead of terminators?  (The language allows you to use them as terminators in DSL source, but that doesn't mean we have to use that in the pretty-printed output as well.)

src/graph.rs
dcreager: The other will be `Debug::fmt`.  The `write!` formulation _might_ get translated into an efficient call; I'm not quite sure how the macro is implemented.  I think you can simplify what you have to:

``` rust
fmt::Display::fmt(value, f)
```

src/graph.rs
```diff
@@ -373,3 +406,35 @@ impl From<GraphNodeRef> for Value {
...
+        struct Impl<'a, 'tree, T: DisplayWithGraph>(&'a T, &'a Graph<'tree>);
+
+        impl<'a, 'tree, T: DisplayWithGraph> fmt::Display for Impl<'a, 'tree, T> {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                self.0.fmt(f, self.1)
+            }
+        }
+
+        Box::new(Impl(self, graph))
+    }
```
dcreager, line 439: Normally I'd suggest that we find a way to avoid the `Box` here, since that means a heap allocation.  If this were a standalone function, and not inside of a trait, you could have the return type be `impl fmt::Display + 'a`, which would avoid the allocation.  But I'm guessing you want to keep this as a method in the trait so that you can call it as `value.display_with` instead of `display_with(value`.

So I see two options:

1. Keep it as `Box`, because this isn't part of the hot path of anything.

2. Move `display_with` outside of the trait, even though it changes how you'd call it.  (That's what I did with the equivalent over in [`stack-graphs`](https://github.com/github/stack-graphs/blob/8f95e1a66db7a1ae9272ba71da1ff33781995f78/src/paths.rs#L78-L92)).

src/graph.rs
```diff
@@ -354,6 +302,78 @@ impl From<String> for Value {
     }
 }
 
+impl fmt::Display for Value {
```
dcreager, line 305: Actually, as a separate question from the comments below about the implementation of this trait — do we _need_ to implement `Display` for `Value`?  Are there places where we don't have access to a `Graph`, and therefore can't call `value.display_with` instead?

dcreager (commented):
Looking good!  Some initial comments

src/graph.rs
hendrikvanantwerpen: This is the Debug formatting, right? Is that something we defined, or does it just do what we want in this case?

src/graph.rs
hendrikvanantwerpen: In principle we always ahve the graph. But, there are two places where I couldn't get it to work because of borrowing:
- https://github.com/tree-sitter/tree-sitter-graph/blob/improve-error-messages/src/execution.rs#L350-L351
- https://github.com/tree-sitter/tree-sitter-graph/blob/improve-error-messages/src/execution.rs#L370-L371
If we can fix that, we could drop Display impl altogether.

dcreager: Ah yes!  Just played around with it locally, and this pattern works:

``` rust
        if let Err(_) = exec.graph[source].add_edge(sink) {
            return Err(ExecutionError::DuplicateEdge(format!(
                "({} -> {}) in {}",
                source.display_with(exec.graph), // cannot display_with(exec.graph) because of borrow
                sink.display_with(exec.graph), // cannot display_with(exec.graph) because of borrow
                self.display_with(ctx)
            )));
        }
```

That way you still have a mutable borrow of `exec` in the conditional, but it's discharged before control passes into the body.

dcreager:
``` rust
if exec.graph[source].add_edge(sink).is_err() {
```

would also work for the conditional if you prefer that spelling.

hendrikvanantwerpen: I think I found a way around this, by matching instead of calling `.map_err` in the chain.

hendrikvanantwerpen: I managed. I have removed the Display impl in favor of DisplayWithGraph. If this is too quickly, let me know and I revert it easily.

src/ast.rs
```diff
@@ -181,6 +296,12 @@ impl From<Scan> for Statement {
     }
 }
 
+impl DisplayWithContext for Scan {
+    fn fmt(&self, f: &mut fmt::Formatter, ctx: &Context) -> fmt::Result {
+        write!(f, "scan {} {{ ... }}", self.value.display_with(ctx))
```
dcreager, line 301: :smile_cat:

src/ast.rs
```diff
@@ -196,13 +317,28 @@ impl PartialEq for ScanArm {
     }
 }
 
+impl DisplayWithContext for ScanArm {
+    fn fmt(&self, f: &mut fmt::Formatter, _ctx: &Context) -> fmt::Result {
+        write!(f, "\"{}\" {{ ... }}", self.regex)
```
dcreager, line 322:
```suggestion
        write!(f, "{:?} {{ ... }}", self.regex.as_str())
```

to automatically get quoting/escaping of the regex

src/ast.rs
hendrikvanantwerpen: This is not an omission :). My thinking is that we format for error messages, and not for pretty-printing a parsible AST. The scan statements, and arms, would be pretty large. So they'll become very long lines (which still wouldn't be valid syntax), or multi-line (with all formatting problems of multi-line errors), and the extra information will not improve the debugging experience much. I am making sure that the arm and statement where the error originated is also included in the context. And, if we have line numbers in the error as well, it will be easy to find in the source, even without the details of the arms.

An example of a trace that originated in a scan arm:
```
Caused by:
    0: Executing scan "a/b/c.py" { ... }
    1: Matching a/b/c.py with arm "([^/]+/)+([^/]*).py$" { ... }
    2: Executing set x = $2
    3: Undefined variable x
```

PR #14, opened 2021-11-08
Add support for nested function calls

tausbn:
A fairly straightforward implementation. Rather than draining the entire
list of parameters from `exec.function_parameters` when calling a
function, we just drain the last `N` where `N` is the number of
arguments passed to the function.

The added test (hopefully) demonstrates that this works correctly.
Without the changes in this PR, execution would panic when the outer
`replace` call was called without arguments (as these would have been
purged by the second call).


hendrikvanantwerpen (approved):
Thanks for fixing this! (It also saved me some work on finding this issue in the lazy implementation.)

PR #15, opened 2021-11-08
Support capture existence testing and simple control flow

tausbn:
Adds support for `if` statements and a prefix operator `?` that can be used to test for the presence of a given capture.

Thus, a stanza like
```
(module . (expression_statement)? @first) @root
{
  node @root.node
  if ?@first {
    attr (@root.node) first_expr = @first
  }
}
```
will set the `first_expr` attribute only in the case where the first statement in the input is in fact an expression.

In addition to this, this PR also adds functions for the usual boolean operators `and`, `or`, and `not`.

Can be reviewed commit-by-commit.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -451,6 +451,8 @@ pub enum Expression {
     Set(SetComprehension),
     // Syntax nodes
     Capture(Capture),
+    // Existence of capture
+    CaptureExists(CaptureExists),
     // Variables
     Variable(Variable),
     // Functions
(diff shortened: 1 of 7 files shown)


dcreager:
As an fyi @hendrikvanantwerpen has also been working on lazy evaluation in the [`lazy-evaluation` branch](https://github.com/tree-sitter/tree-sitter-graph/tree/lazy-evaluation).  I know that involves some conditional logic under the covers to support evaluating `scan` statements lazily.  We should make sure these two proposals / PRs don't conflict with each other.

tausbn:
> As an fyi @hendrikvanantwerpen has also been working on lazy evaluation in the [`lazy-evaluation` branch](https://github.com/tree-sitter/tree-sitter-graph/tree/lazy-evaluation). I know that involves some conditional logic under the covers to support evaluating `scan` statements lazily. We should make sure these two proposals / PRs don't conflict with each other.

Yep, I'm aware (though that branch doesn't seem to be in a good state for reviewing yet, so I haven't looked at its contents in any detail).

The way I see it, this PR is just a proposal. I'm personally not enthused with the `?@foo` syntax, and so I wouldn't expect to see this merged in its current form (or at all, for that matter). Mostly I just wanted it to spark some discussion about the benefits of adding control flow. (And just implementing my suggestion seemed like the most straightforward way to do so. :slightly_smiling_face:)

hendrikvanantwerpen:
Based on this proposal I have implemented #20, which adds conditionals and list iteration, as well as improves handling of captures of patterns with suffixes such as `?` and `*`. I therefore propose to close this PR in favor of #20.

tausbn:
Agreed! Closing.

PR #16, opened 2021-11-09
Add support for parsing list and set literals

tausbn:
As far as I can tell, there is already support for evaluating these correctly. We were just missing the necessary parts of the parser.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -487,6 +487,8 @@ impl Parser<'_> {
             '@' => self.parse_capture(current_query)?,
             '$' => self.parse_regex_capture()?,
             '(' => self.parse_call(current_query)?,
+            '[' => self.parse_list(current_query)?,
+            '{' => self.parse_set(current_query)?,
             ch if ch.is_ascii_digit() => self.parse_integer_constant()?,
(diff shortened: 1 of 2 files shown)


hendrikvanantwerpen (approved):
This is probably a good idea.

We probably want to add list/set related functions (length/size, nth list element) as well, so that you can really use these properly?

tausbn:
> This is probably a good idea.

I mean, this is syntax that is described in the `tree-sitter-graph` reference, so I figured we'd better add it. :wink: 
 
> We probably want to add list/set related functions (length/size, nth list element) as well, so that you can really use these properly?

That sounds reasonable, though I don't have any immediate use for such functions. I just wanted a convenient way of writing tuples to attributes.

hendrikvanantwerpen:
> I mean, this is syntax that is described in the tree-sitter-graph reference, so I figured we'd better add it. 😉

Makes sense :)

> That sounds reasonable, though I don't have any immediate use for such functions. I just wanted a convenient way of writing tuples to attributes.

Let's merge this as is, then, and think about useful functions when the need arises.
(I'm assuming you can merge this, but otherwise ping me and I'll do it for you.)

PR #17, opened 2021-11-10
Fix `print` statement

tausbn:
We were accidentally consuming the next token whenever `try_next`
returned `Some(x)` where `x != ','`. This obviously caused a lot of
problems further down the line.

To fix this, I changed the code to use `peek` instead.

This made the `try_next` function unused, so to remove the warning I
also added an `#[allow(dead_code)]` (just in case it becomes useful
later on).

---
Also ensures that strings are printed verbatim (and without escaping),
to make the printed output a bit nicer.

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -438,14 +438,14 @@ impl Scan {
 impl Print {
     fn execute(&self, exec: &mut ExecutionContext) -> Result<(), ExecutionError> {
         for value in &self.values {
-            let value = value.evaluate(exec)?;
-            if let Value::String(s) = value {
-                print!("{}", s);
+            if let Expression::StringConstant(expr) = value {
+                eprint!("{}", expr.value);
             } else {
-                print!("{}", value.display_with(exec.graph));
+                let value = value.evaluate(exec)?;
+                eprint!("{}", value.display_with(exec.graph));
             }
         }
-        println!();
+        eprintln!();
         Ok(())
     }
 }
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -439,9 +439,13 @@ impl Print {
     fn execute(&self, exec: &mut ExecutionContext) -> Result<(), ExecutionError> {
         for value in &self.values {
             let value = value.evaluate(exec)?;
-            print!("{}", value.display_with(exec.graph));
+            if let Value::String(s) = value {
+                print!("{}", s);
+            } else {
+                print!("{}", value.display_with(exec.graph));
+            }
         }
-        println!("");
+        println!();
         Ok(())
     }
 }
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -162,6 +162,7 @@ impl Parser<'_> {
         Ok(ch)
     }
 
+    #[allow(dead_code)]
     fn try_next(&mut self) -> Option<char> {
         self.next().ok()
     }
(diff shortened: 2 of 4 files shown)


tausbn:
Sorry about the force push. The conflict between this and #16 was somewhat gnarly to untangle.

src/parser.rs
```diff
@@ -162,6 +162,7 @@ impl Parser<'_> {
         Ok(ch)
     }
 
+    #[allow(dead_code)]
     fn try_next(&mut self) -> Option<char> {
         self.next().ok()
     }
```
hendrikvanantwerpen, line 168: I think it is okay to just remove this.

src/execution.rs
```diff
@@ -439,9 +439,13 @@ impl Print {
     fn execute(&self, exec: &mut ExecutionContext) -> Result<(), ExecutionError> {
         for value in &self.values {
             let value = value.evaluate(exec)?;
-            print!("{}", value.display_with(exec.graph));
+            if let Value::String(s) = value {
+                print!("{}", s);
+            } else {
+                print!("{}", value.display_with(exec.graph));
+            }
```
hendrikvanantwerpen, line 446: I am not a big fan of obscuring details about the values being printed (e.g., `1` and `"1"` become indistinguishable. When debugging I prefer to clearly see what kind of value is there. But from the test I see what pattern you use for debugging, and it makes sense that you don't want the quoted version of the literals you add. I suggest changing the logic slightly, to only omit the quotes for string literals, not all string values.

```suggestion
            if let Expression::StringConstant(expr) = value {
                print!("{}", expr.value);
            } else {
                let value = value.evaluate(exec)?;
                print!("{}", value.display_with(exec.graph));
            }
```

hendrikvanantwerpen (commented):
I left a few comments below.

src/execution.rs
tausbn: Ah, this is an excellent point (and suggestion).

Incidentally, would you have any objections to changing `print!` to `eprint!` in order to make the printed statements appear on `stderr` instead? The way the graph is output currently, anything output using `print` has the potential to mess up the consumers of the graph output.

tausbn: Hm. For some reason I'm not allowed to apply your suggestion here directly (GitHub complains about suggestions added to deleted lines. I'll see if I can fix it (and if not I'll just perform the change manually).

hendrikvanantwerpen: No, using `eprint` sounds like a good idea. In the future we may want to have some flags controlling output channel and format, but for now this would make the current behavior more robust indeed.

hendrikvanantwerpen (approved):
One question got lost from my previous review, so I've added that below. Otherwise, this looks good!

src/parser.rs
```diff
@@ -396,10 +392,13 @@ impl Parser<'_> {
         } else if keyword == self.print_keyword {
             let mut values = vec![self.parse_expression(current_query)?];
             self.consume_whitespace();
-            while self.try_next() == Some(',') {
+            while self.peek()? == ',' {
```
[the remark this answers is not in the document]
tausbn: Ah, I was just about to bring this up myself (having seen how multiple attributes are parsed in the meantime). 

I think ultimately an error in `peek()` will be fatal anyway, since that can only happen when there is an unexpected EOF, and since `print` statements can only appear inside stanzas, they cannot appear immediately before EOF.

However, from that point of view, the parse error is _not_ with the `print` statement itself, and so it makes more sense to build the value and make the error appear when we try to close the surrounding stanza.

PR #18, opened 2021-11-11
Reject nullable regular expressions

hendrikvanantwerpen:
Regular expressions that match empty strings cause `scan` to loop indefinitely. This PR fixes that problem by skipping empty matches.


src/execution.rs
```diff
@@ -396,7 +396,9 @@ impl Scan {
             for (index, arm) in self.arms.iter().enumerate() {
                 let captures = arm.regex.captures(&match_string[i..]);
                 if let Some(captures) = captures {
-                    matches.push((captures, index));
+                    if !captures.get(0).unwrap().range().is_empty() {
```
tausbn, line 399: Hmm... Is there a situation where you would want to have (potentially) an empty match (which will now be skipped), or would it perhaps make more sense to throw an error if an empty match is ever encountered?

I can't off the top of my head think of a situation where allowing empty matches is actually beneficial, but admittedly I haven't used `scan` in this way much.

hendrikvanantwerpen: That is a good point. I don't think allowing an empty match is ever useful. If it was intended, you can just lift those statements out of the scan, as it is unconditional. Throwing an error helps pointing out the issue to the user, probably improving the overall experience.

hendrikvanantwerpen:
I have changed this fix to reject nullable regular expression at parse time of the query file.

src/parser.rs
```diff
@@ -414,7 +416,15 @@ impl Parser<'_> {
                 let pattern_location = self.location;
                 let pattern = self.parse_string()?;
                 let regex = match Regex::new(&pattern) {
-                    Ok(regex) => regex,
+                    Ok(regex) => {
+                        if let Some(_) = regex.captures("") {
```
tausbn, line 420: This won't prevent _all_ empty matches, but should suffice for the vast majority, so I think it's probably fine.

(For instance, the regex `"^\b"` matches an empty word boundary at the beginning of the string. This _won't_ match the empty string, but will produce an empty match on, say, the string `"a"`)

src/execution.rs
```diff
@@ -396,9 +396,10 @@ impl Scan {
             for (index, arm) in self.arms.iter().enumerate() {
                 let captures = arm.regex.captures(&match_string[i..]);
                 if let Some(captures) = captures {
-                    if !captures.get(0).unwrap().range().is_empty() {
-                        matches.push((captures, index));
+                    if captures.get(0).unwrap().range().is_empty() {
+                        panic!("Empty match for regular expression /{}/", arm.regex);
```
tausbn, line 400: Using `/` as the delimiters seems a bit nonstandard, given that the syntax used in an actual piece of code doesn't use these delimiters. 

As long as it's unambiguous, though, I think it's fine.

tests/it/execution.rs
```diff
@@ -196,8 +228,8 @@ fn can_nest_function_calls() {
 }
 
 #[test]
-fn skips_empty_matches() {
-    check_execution(
+fn cannot_use_nullable_regex() {
+    fail_parse(
```
tausbn, line 232: Is there a reason why this test of the parser doesn't live in `tests/it/parser.rs`?

tausbn (changes requested):
A few comments, but overall I think this looks good. :+1:

src/execution.rs
hendrikvanantwerpen: It was already used in the format string of an existing exception. It is maybe even nicer if this was used as the syntax as well, instead of quotes, to be more explicit about it being a regex instead of a simple string.

tests/it/execution.rs
hendrikvanantwerpen: You are right, this should not be here. Let me fix that.

src/parser.rs
hendrikvanantwerpen: Interesting example :), I did not think about that. For this the check in `evaluate` is still useful then. I've adapted the execution test to cover this case.

hendrikvanantwerpen: Unfortunately Rust regular expressions don't tell you if they are nullable, so we'll have to live with most caught statically, and some dynamically.

tausbn (approved):
Looks good to me!

src/execution.rs
tausbn: Good point!

hendrikvanantwerpen:
Thanks for the comments, @tausbn. Those were very useful.

tausbn:
You're welcome! And likewise thank you for reviewing my PRs. :slightly_smiling_face:

PR #20, opened 2021-11-24
Add control statements and improve handling of pattern suffixes

hendrikvanantwerpen:
This PR introduces control statements for conditionals and iteration,
and improves the handling of patterns with suffixes indicating list or
optional outputs.

Can be reviewed commit by commit.

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -19,12 +19,14 @@ use crate::ast::AddGraphNodeAttribute;
 use crate::ast::Assign;
 use crate::ast::Call;
 use crate::ast::Capture;
+use crate::ast::Condition;
 use crate::ast::CreateEdge;
 use crate::ast::CreateGraphNode;
 use crate::ast::DeclareImmutable;
 use crate::ast::DeclareMutable;
 use crate::ast::Expression;
 use crate::ast::File;
+use crate::ast::If;
 use crate::ast::IntegerConstant;
 use crate::ast::ListComprehension;
 use crate::ast::Print;
@@ -227,6 +229,7 @@ impl Statement {
             Statement::AddEdgeAttribute(statement) => statement.execute(exec),
             Statement::Scan(statement) => statement.execute(exec),
             Statement::Print(statement) => statement.execute(exec),
+            Statement::If(statement) => statement.execute(exec),
         }
     }
 }
@@ -411,6 +414,56 @@ impl Print {
         }
         eprintln!();
         Ok(())
+    }
+}
+
+impl If {
+    fn execute(&self, exec: &mut ExecutionContext) -> Result<(), ExecutionError> {
+        for arm in &self.arms {
+            let mut result = true;
+            for condition in &arm.conditions {
+                result &= condition.test(exec)?;
+            }
+            if result {
+                let mut arm_locals = VariableMap::new_child(exec.locals);
+                let mut arm_exec = ExecutionContext {
+                    ctx: exec.ctx,
+                    source: exec.source,
+                    graph: exec.graph,
+                    functions: exec.functions,
+                    globals: exec.globals,
+                    locals: &mut arm_locals,
+                    scoped: exec.scoped,
+                    current_regex_captures: exec.current_regex_captures,
+                    function_parameters: exec.function_parameters,
+                    mat: exec.mat,
+                };
+                for stmt in &arm.statements {
+                    stmt.execute(&mut arm_exec)?;
+                }
+                break;
+            }
+        }
+        Ok(())
+    }
+}
+
+impl Condition {
+    fn test(&self, exec: &mut ExecutionContext) -> Result<bool, ExecutionError> {
+        let mut result = true;
+        match self {
+            Condition::Some(captures) => {
+                for capture in captures {
+                    result &= !capture.evaluate(exec)?.is_null();
+                }
+            }
+            Condition::None(captures) => {
+                for capture in captures {
+                    result &= capture.evaluate(exec)?.is_null();
+                }
+            }
+        }
+        Ok(result)
     }
 }
 
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -26,6 +26,7 @@ use crate::ast::DeclareImmutable;
 use crate::ast::DeclareMutable;
 use crate::ast::Expression;
 use crate::ast::File;
+use crate::ast::ForIn;
 use crate::ast::If;
 use crate::ast::IntegerConstant;
 use crate::ast::ListComprehension;
@@ -230,6 +231,7 @@ impl Statement {
             Statement::Scan(statement) => statement.execute(exec),
             Statement::Print(statement) => statement.execute(exec),
             Statement::If(statement) => statement.execute(exec),
+            Statement::ForIn(statement) => statement.execute(exec),
         }
     }
 }
@@ -464,6 +466,33 @@ impl Condition {
             }
         }
         Ok(result)
+    }
+}
+
+impl ForIn {
+    fn execute(&self, exec: &mut ExecutionContext) -> Result<(), ExecutionError> {
+        let values = self.capture.evaluate(exec)?.into_list(exec.graph)?;
+        let mut loop_locals = VariableMap::new_child(exec.locals);
+        for value in values {
+            loop_locals.clear();
+            let mut loop_exec = ExecutionContext {
+                ctx: exec.ctx,
+                source: exec.source,
+                graph: exec.graph,
+                functions: exec.functions,
+                globals: exec.globals,
+                locals: &mut loop_locals,
+                scoped: exec.scoped,
+                current_regex_captures: exec.current_regex_captures,
+                function_parameters: exec.function_parameters,
+                mat: exec.mat,
+            };
+            self.variable.add(&mut loop_exec, value, false)?;
+            for stmt in &self.statements {
+                stmt.execute(&mut loop_exec)?;
+            }
+        }
+        Ok(())
     }
 }
 
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -33,6 +33,7 @@ use crate::ast::ListComprehension;
 use crate::ast::Print;
 use crate::ast::RegexCapture;
 use crate::ast::Scan;
+use crate::ast::ScanExpression;
 use crate::ast::ScopedVariable;
 use crate::ast::SetComprehension;
 use crate::ast::Stanza;
@@ -404,6 +405,17 @@ impl Scan {
     }
 }
 
+impl ScanExpression {
+    fn evaluate(&self, exec: &mut ExecutionContext) -> Result<Value, ExecutionError> {
+        match self {
+            Self::StringConstant(expr) => expr.evaluate(exec),
+            Self::Capture(expr) => expr.evaluate(exec),
+            Self::Variable(expr) => expr.get_global(exec).map(|v| v.clone()),
+            Self::RegexCapture(expr) => expr.evaluate(exec),
+        }
+    }
+}
+
 impl Print {
     fn execute(&self, exec: &mut ExecutionContext) -> Result<(), ExecutionError> {
         for value in &self.values {
@@ -701,6 +713,12 @@ impl ScopedVariable {
 }
 
 impl UnscopedVariable {
+    fn get_global<'a>(&self, exec: &'a mut ExecutionContext) -> Result<&'a Value, ExecutionError> {
+        exec.globals.get(self.name).ok_or_else(|| {
+            ExecutionError::UndefinedVariable(format!("{}", self.display_with(exec.ctx)))
+        })
+    }
+
     fn get<'a>(&self, exec: &'a mut ExecutionContext) -> Result<&'a Value, ExecutionError> {
         if let Some(value) = exec.globals.get(self.name) {
             Some(value)
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -413,6 +413,13 @@ pub enum Value {
 }
 
 impl Value {
+    /// Check if this value is null
+    pub fn is_null(self) -> bool {
+        match self {
+            Value::Null => true,
+            _ => false,
+        }
+    }
     /// Coerces this value into a boolean, returning an error if it's some other type of value.
     pub fn into_bool(self, graph: &Graph) -> Result<bool, ExecutionError> {
         match self {
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -36,6 +36,8 @@ pub enum ParseError {
     ExpectedVariable(Location),
     #[error("Expected unscoped variable at {0}")]
     ExpectedUnscopedVariable(Location),
+    #[error("Expected list capture {0}")]
+    ExpectedListCapture(Location),
     #[error("Expected optional capture {0}")]
     ExpectedOptionalCapture(Location),
     #[error("Invalid regular expression /{0}/ at {1}")]
@@ -94,6 +96,7 @@ struct Parser<'a> {
     attr_keyword: Identifier,
     edge_keyword: Identifier,
     false_keyword: Identifier,
+    for_keyword: Identifier,
     if_keyword: Identifier,
     let_keyword: Identifier,
     node_keyword: Identifier,
@@ -119,6 +122,7 @@ impl<'a> Parser<'a> {
         let attr_keyword = ctx.add_identifier("attr");
         let edge_keyword = ctx.add_identifier("edge");
         let false_keyword = ctx.add_identifier("false");
+        let for_keyword = ctx.add_identifier("for");
         let if_keyword = ctx.add_identifier("if");
         let let_keyword = ctx.add_identifier("let");
         let node_keyword = ctx.add_identifier("node");
@@ -137,6 +141,7 @@ impl<'a> Parser<'a> {
             attr_keyword,
             edge_keyword,
             false_keyword,
+            for_keyword,
             let_keyword,
             if_keyword,
             node_keyword,
@@ -503,6 +508,31 @@ impl Parser<'_> {
 
             Ok(ast::If {
                 arms,
+                location: keyword_location,
+            }
+            .into())
+        } else if keyword == self.for_keyword {
+            self.consume_whitespace();
+            let variable = match self.parse_variable(current_query)? {
+                ast::Variable::Unscoped(variable) => Ok(variable),
+                ast::Variable::Scoped(variable) => {
+                    Err(ParseError::ExpectedUnscopedVariable(variable.location))
+                }
+            }?;
+            self.consume_whitespace();
+            self.consume_token("in")?;
+            self.consume_whitespace();
+            let capture_location = self.location;
+            let capture = self.parse_capture(current_query)?;
+            if capture.quantifier != ZeroOrMore && capture.quantifier != OneOrMore {
+                return Err(ParseError::ExpectedListCapture(capture_location));
+            }
+            self.consume_whitespace();
+            let statements = self.parse_statements(current_query)?;
+            Ok(ast::ForIn {
+                variable,
+                capture,
+                statements,
                 location: keyword_location,
             }
             .into())
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -11,6 +11,7 @@ use std::str::Chars;
 
 use regex::Regex;
 use thiserror::Error;
+use tree_sitter::CaptureQuantifier::*;
 use tree_sitter::Language;
 use tree_sitter::Query;
 use tree_sitter::QueryError;
@@ -33,6 +34,10 @@ pub enum ParseError {
     ExpectedToken(&'static str, Location),
     #[error("Expected variable name at {0}")]
     ExpectedVariable(Location),
+    #[error("Expected unscoped variable at {0}")]
+    ExpectedUnscopedVariable(Location),
+    #[error("Expected optional capture {0}")]
+    ExpectedOptionalCapture(Location),
     #[error("Invalid regular expression /{0}/ at {1}")]
     InvalidRegex(String, Location),
     #[error("Nullable regular expression /{0}/ at {1}")]
@@ -89,6 +94,7 @@ struct Parser<'a> {
     attr_keyword: Identifier,
     edge_keyword: Identifier,
     false_keyword: Identifier,
+    if_keyword: Identifier,
     let_keyword: Identifier,
     node_keyword: Identifier,
     null_keyword: Identifier,
@@ -113,6 +119,7 @@ impl<'a> Parser<'a> {
         let attr_keyword = ctx.add_identifier("attr");
         let edge_keyword = ctx.add_identifier("edge");
         let false_keyword = ctx.add_identifier("false");
+        let if_keyword = ctx.add_identifier("if");
         let let_keyword = ctx.add_identifier("let");
         let node_keyword = ctx.add_identifier("node");
         let null_keyword = ctx.add_identifier("null");
@@ -131,6 +138,7 @@ impl<'a> Parser<'a> {
             edge_keyword,
             false_keyword,
             let_keyword,
+            if_keyword,
             node_keyword,
             null_keyword,
             print_keyword,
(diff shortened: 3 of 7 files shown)


src/execution.rs
```diff
@@ -59,6 +59,9 @@ impl File {
         globals: &mut Variables,
     ) -> Result<Graph<'tree>, ExecutionError> {
         let mut graph = Graph::new();
+        if tree.root_node().has_error() {
+            return Err(ExecutionError::ParseTreeHasErrors);
+        }
```
tausbn, line 67: Do I understand correctly that this means that if a file has but a single syntax error, it will get ignored completely?

From my perspective (and use case) that seems rather severe. I would rather continue with the execution regardless (since it may very well be that the syntax error has a limited impact on the rest of the file).

At the very least, we could make this behaviour dependent on a command-line flag.

tests/it/parser.rs
```diff
@@ -167,6 +174,7 @@ fn can_parse_blocks() {
                     scope: Box::new(
                         Capture {
                             index: 1,
+                            suffix: QueryCaptureSuffix::One,
                             name: cap2
                         }
                         .into()
```
tausbn, line 180: It feels slightly odd to not see any tests with the `*`, `+`, and `?` suffixes.

src/functions.rs
```diff
@@ -101,20 +101,39 @@ impl Functions {
...
-        functions.add(ctx.add_identifier("start-row"), stdlib::StartRow);
-        functions.add(ctx.add_identifier("start-column"), stdlib::StartColumn);
-        functions.add(ctx.add_identifier("end-row"), stdlib::EndRow);
-        functions.add(ctx.add_identifier("end-column"), stdlib::EndColumn);
-        functions.add(ctx.add_identifier("node-type"), stdlib::NodeType);
+        // general functions
+        functions.add(ctx.add_identifier("is-null"), stdlib::IsNull);
+        // tree functions
+        functions.add(
+            ctx.add_identifier("child-index"),
```
tausbn, line 108: One thing I've been wondering (and now seems like a good time to bring it up) is whether we should rename this to `named-child-index`.

I ended up requiring a function that got the index of _any_ child (including unnamed children), and had to call it `unnamed-child-index` because the more sensible name was taken already. Worse, `unnamed-child-index` is not even the right name (but I couldn't think of another, better name) because it also includes named children (by virtue of simply dispatching to the `tree-sitter` `children` method).

src/functions.rs
```diff
@@ -158,205 +177,298 @@ pub mod stdlib {
...
+                _source: &str,
+                parameters: &mut dyn Parameters,
+            ) -> Result<Value, ExecutionError> {
+                let mut result = true;
+                while let Ok(parameter) = parameters.param() {
+                    result &= parameter.into_bool(graph)?;
+                }
+                Ok(result.into())
+            }
+        }
```
tausbn, line 407: I have the strangest feeling of déjà vu... :thinking:

src/execution.rs
```diff
@@ -484,6 +492,23 @@ impl Conditional {
...
+        let values = self.values.evaluate(exec)?;
+        let values = values.into_list(exec.graph)?;
+        for value in values {
+            exec.locals.add_immutable(self.name, value).map_err(|_| {
+                ExecutionError::DuplicateVariable(format!("{}", self.name.display_with(exec.ctx)))
+            })?;
+            for stmt in &self.statements {
+                stmt.execute(exec)?;
+            }
+            exec.locals.remove(self.name);
```
tausbn, line 506: I am happy to see that the loop variable doesn't persist after the loop is done. :+1:

tausbn (changes requested):
Really nice! (Especially the handling of captures!) 

I have a few comments, but apart from that I'm giddy with excitement for getting access to these constructs. :heart_eyes:

src/graph.rs
```diff
@@ -255,6 +255,17 @@ pub enum Value {
 }
 
 impl Value {
+    /// Coerces this value into a boolean, returning an error if it's some other type of value.
+    pub fn into_bool(self, graph: &Graph) -> Result<bool, ExecutionError> {
+        match self {
+            Value::Boolean(value) => Ok(value),
```
tausbn, line 261: I'm sorely tempted to suggest allowing coercing `#null` into `False` and any `tree-sitter` syntax node into `True`. This would make it a bit more ergonomic when checking for the presence of captures, and avoid a whole bunch of `(is-null ...)` calls.

src/execution.rs
```diff
@@ -689,3 +686,39 @@ impl UnscopedVariable {
...
+            let syntax_node = graph.add_syntax_node(nodes[0]);
+            syntax_node.into()
+        }
+        QueryCaptureSuffix::ZeroOrMore | QueryCaptureSuffix::OneOrMore => {
+            let syntax_nodes = nodes
+                .iter()
+                .map(|n| graph.add_syntax_node(n.clone()).into())
+                .collect::<Vec<Value>>();
+            syntax_nodes.into()
+        }
```
tausbn, line 755: What happens if an outer `+` or `*` capture contains inner captures?
E.g. something like
```
[
    (foo (_) @bar)
    (baz (_) @quux)
]+ @foo_or_baz
```
Would `@bar`/`@quux` just fail to to have a value at all? Or would it end up with the value corresponding to the last capture of `@foo_or_baz`?

hendrikvanantwerpen: Good question. I expect it will have the last value, but I haven't tested that yet.

src/functions.rs
hendrikvanantwerpen: I think that is a good idea. It describes its meaning better, and is also more consistent with `named-child-count`.

src/execution.rs
hendrikvanantwerpen: :)

src/functions.rs
hendrikvanantwerpen: Not sure what you're hinting at here...

tausbn: Nothing important. Just that these are the exact same functions I implemented in #15. :laughing:

src/execution.rs
hendrikvanantwerpen: TLDR: If this check is a problem for you right now, I'll take it out. I may have an idea how to do this more fine-grained, but that'll be a different PR.

> Do I understand correctly that this means that if a file has but a single syntax error, it will get ignored completely?

Yes.

> From my perspective (and use case) that seems rather severe. I would rather continue with the execution regardless

I agree that this would be more desirable. The reason I did it like this (for now) is because of similar problems we had with `tree-sitter-code-nav`. There executing stanzas on trees with errors would sometimes result in hard panics---also not ideal for a library.

I haven't had time to investigate exactly if we can be more precise while still avoiding the errors. Let's assume that query matching is somewhat sane, so a query only matches if all named nodes in the pattern are actually there in the tree. That means we can only refer to error nodes in leaf, wildcard captures. If the capture is an error node itself, using the capture will probably always result in an error:
1. Accessing the source text might be a problem. I think this was the main reason for the panics in `tree-sitter-code-nav`, when there were parse errors for identifier names.
2. It is (probably) missing expected scoped variables defined on it. The capture could also have error nodes deeper in the subtree, which may or may not prevent other stanzas to fire, which may also result in errors because of missing scoped variables.

Okay, now that I am writing this up, one strategy that may work is introducing a special exception that indicates we hit an error node during evaluation. We maybe only have to throw it when we convert a syntax reference to the actual node---thereby guaranteeing that if we get a node, it is never an error node (although it may have error nodes as descendants). Instead of always propagating this exception to the top, we can absorb it at a statement level. So, any statement that involves an error node will be skipped. However, this this not solve the missing scoped variables completely, because values may not be set if they hit an error node. If we can also mark (scoped) variables as error-like if their value could not be computed because of an error node, then we have enough to avoid missing (scoped) variable errors, I think.

I think this is worth trying, but I'll do that in a different PR. If this check is problematic for you atm, I'll take it out. Let me know.

src/graph.rs
hendrikvanantwerpen: I understand why it would be nice. I think a decision on that depends on what we expect/hope to do in the future in terms of static checking of DSL files. Such convenience features, which are easy to support in an untyped language, can really hurt when you later want to add checking. That is why I am usually quite conservative about them.

On the other hand, since optional captures are an integral part of the queries, this warrants special syntax, I suppose. Would a `?` postfix operator satisfy your temptation?

tausbn: I agree about this potentially harming the kind of static analysis we can do. I tried my best to think of a case where this (i.e. a capture silently being coerced into a boolean) might introduce a subtle bug, but I couldn't come up with a convincing example of this.

I think the postfix `?` syntax is likely sufficiently ergonomic.

Speaking of type safety (and being conservative), we should consider using a special `Value::NoCapture` constant instead of `Value::Null`. Then we could limit the `?` operator to just captures until such a time as we find a more general need for it. (I'm sure a nullness check _could_ have other uses, but I can't think of any at the moment.)

tausbn: I agree about this potentially harming the kind of static analysis we can do. I tried my best to think of a case where this (i.e. a capture silently being coerced into a boolean) might introduce a subtle bug, but I couldn't come up with a convincing example of this.

I think the postfix `?` syntax is likely sufficiently ergonomic.

Speaking of type safety (and being conservative), we should consider using a special `Value::NoCapture` constant instead of `Value::Null`. Then we could limit the `?` operator to just captures until such a time as we find a more general need for it. (I'm sure a nullness check _could_ have other uses, but I can't think of any at the moment.)

src/execution.rs
tausbn: Hmm... You're right about things likely failing due to missing scoped variables.

In that case, I think it's probably better to include this change.

Truth be told, I'm not entirely sure how `ERROR` nodes work. I had assumed that they were just a slightly odd kind of named node (with a location and potentially with children). However, seeing as it isn't possible to query for these nodes, I'm not too sure how accurate this mental image is.

Ideally, I would want to be able to query for these `ERROR` nodes and also capture and assign scoped variables to them. That way, I can just treat them as another kind of AST node, link them up appropriately (without running into missing scoped variables), and then recognise and handle them appropriately in later stages of the analysis.

tests/it/execution.rs
```diff
@@ -343,3 +343,149 @@ fn can_execute_for_in_empty_list() {
...
+        "#},
+        indoc! {r#"
+          node 0
+            val: 2
+        "#},
+    );
+}
+
+#[test]
+fn can_execute_plus_capture_empty() {
```
tausbn, line 421: I mean, is this _really_ showing it can execute a `+` when the capture is empty? :slightly_smiling_face: 

It looks more like it's `does_not_execute_plus_capture_empty`.

src/functions.rs
```diff
@@ -203,9 +203,9 @@ pub mod stdlib {
 
         /// The implementation of the standard [`child-index`][`crate::reference::functions#child-index`]
```
tausbn, line 204: This should also refer to `named-child-index`.

(I would have made a suggested change, but it's outside the lines that have been touched by this PR, so the interface doesn't allow me to.)

tausbn (approved):
A few comments, otherwise this looks good to me! :+1:

src/functions.rs
hendrikvanantwerpen: Good catch!

tests/it/execution.rs
hendrikvanantwerpen: Fair point. In that case I'll drop this test, as it really just tests for `tree-sitter` behavior then.

src/graph.rs
hendrikvanantwerpen: > I tried my best to think of a case where this (i.e. a capture silently being coerced into a boolean) might introduce a subtle bug

I also could not come up with one, so it's more an argument of potential type checker complexity than correctness problems.

I'll leave it as this for the moment, and poke @dcreager to discuss some guidelines for how to approach these decisions in the future.

hendrikvanantwerpen: After thinking about your comments some more, I have developed more concerns for a future typing approach. Assume optional values to have a `Option<T>` type. We now use `_? : Option<T> -> bool` to test if it is zero. But we use the `Option<T>` in a `T` position without an explicit operation to without explicitly converting `Option<T> -> T`. So there is still implicit coercion (or some form of flow typing) involved here.

Assuming we want option to be properly type checked, a "simple" type checker would require explicit operations for `is-null` as well as `unwrap`. Any setup where any one of these is implicit would require flow typing, coercion, (maybe other options?), or just not representing nullability in the types.

tausbn: I guess ideally we would have that the following code
```
if @capture? { ... }
```
behaves as the following pseudo-Rust-code.
```rust
if let Some(@capture) = @capture { ... }
```
In other words, we match _and rebind_ `@capture` based on whether it's `Some` or `None`. That way, we don't have to explicitly `unwrap` it inside the `if`.

That would make the coercion explicit, though it would also mean that there's no way to do complex boolean tests with captures. :disappointed: 

I see your point about type checking, though. I'm assuming "flow typing" refers to the fact that it's safe to access `@capture` if we're inside a conditional where the condition implies `@capture` cannot be empty. The trouble here, of course, is that it may be difficult to determine whether this implication is actually true.

On its own, I don't see how `unwrap` helps, since that surely just punts the checking to runtime instead. Likewise for the implicit coercion.

Of course, there is also the option (which may be the same as your final option) that is more in line with what I suggested in #15: make the `?` operator something special that inspects the local environment for the existence of a given capture and returns a boolean based on whether it's there or not. Then `(_)? @capture` can behave as it does currently, either setting `@capture` to the captured syntax node, or _not_ setting it if the match is empty. No option types would be needed then.

I guess that's also undesirable, though. Ultimately, we would want to know that all captures either _must_ exist, or if they don't, that this has been accounted for through control-flow.

This all makes me think that putting the control flow _inside_ the stanzas is perhaps a bad idea (apart from `scan` which is just weird, but in a very local way). Instead, maybe we should push this checking of capture existence to the top level of the stanza -- or rather, allow multiple, guarded sets of actions for a single query. 

So, something like
```
(foo
  bar: (_)? @bar
  baz: (_)? @baz
  quux: (_) @quux
) @foo
{ ; Default stanza, always run
  ... ; we can freely refer to @foo and @quux in here
}
(@bar & @baz) { ; only run if @bar and @baz are both defined
  ... ; we can freely refer to @foo, @bar, @baz, and @quux in here
}
(@bar & !baz) { ; only run if @bar is defined and @baz is not
  ... ; we can freely refer to @foo, @bar, and @quux in here
}
...
```
Crucially, we would only allow boolean expressions over captures, so figuring out whether something is implied should be straightforward (and in practice, I don't expect these expressions to be terribly complicated). (In fact, unless I'm mistaken, these "guards" would essentially just reduce to conjunctions of possibly negated captures.)

I apologise for turning this into a sort of meandering comment. To be honest, I would be fine with just picking an option (like this PR) for now, and then later on fixing things up to be more robust.

hendrikvanantwerpen: > I apologise for turning this into a sort of meandering comment.

Not at all, you raise some interesting points!

> To be honest, I would be fine with just picking an option (like this PR) for now, and then later on fixing things up to be more robust.

I think that is a good idea.

hendrikvanantwerpen:
I was not careful enough and messed up part of the history I already pushed---sorry for the force push.

Summary of the changes since your last review:
- `child-index` -> `named-child-index`
- The `cond` syntax was dropped in favor of adding `elif` and `else` to `if` and having only one syntax for conditionals instead of two.
- Updated documentation for changed and updated things.

src/parser.rs
```diff
@@ -445,6 +448,58 @@ impl Parser<'_> {
...
+            let location = keyword_location;
+            let condition = self.parse_expression(current_query)?;
+            self.consume_whitespace();
+            let statements = self.parse_statements(current_query)?;
+            self.consume_whitespace();
+            arms.push(ast::ConditionalArm {
+                condition,
+                statements,
+                location,
+            });
```
tausbn, line 467: Would it perhaps make sense to factor this out into its own function? The repetition in all three cases (`if`/`elif`/`else`) is a bit much. (Though of course the `else` branch is special.)

tausbn (approved):
One minor comment, but otherwise this (still) looks good to me! :slightly_smiling_face:

src/execution.rs
```diff
@@ -458,6 +479,37 @@ impl Print {
...
 }
 
+impl Conditional {
+    fn execute(&self, exec: &mut ExecutionContext) -> Result<(), ExecutionError> {
+        for arm in &self.arms {
+            if arm.condition.evaluate(exec)?.into_bool(exec.graph)? {
+                for stmt in &arm.statements {
+                    stmt.execute(exec)?;
+                }
+                return Ok(());
```
tausbn, line 489: Unless I'm mistaken, this could also just be `break` (which might make it more clear that only one arm is ever executed).

src/parser.rs
hendrikvanantwerpen: I've tried a few times to reduce the duplication, but so far without success. The small differences always get in the way of abstracting a substantial part of those blocks.

tausbn: Yeah that's what I figured.

At any rate, I'm planning to replace the parser with a custom `tree-sitter` grammar at some point anyway, so it doesn't really matter how slick the current parser is.

src/execution.rs
tausbn: One possibility (though I don't know how easy it would be be to implement) would be to also produce lists of captures for `@bar` and `@quux`, but use `#null` (or whatever we use to indicate "no capture") for those indices where there was no capture.

I don't know how useful this would be, though. To tie, say `@bar` together with `@foo_or_baz`, you would then have to zip together these two lists.

hendrikvanantwerpen: Feedback on tree-sitter/tree-sitter#1504 noted that my approach was too simplistic, because it doesn't account for these nesting scenarios. I am revisiting that to make sure that is handled better.

`tree-sitter` outputs a list of captured nodes, indicating to which capture the node belongs. The output will be simpler than a full matching tree, so reconstructing such rich output that allows us to match nodes of one capture to nodes of another would require (probably invasive) tree-sitter changes. But I also think that this would over-complicate things, without clear benefits.

Consider the following example:

    (foo (_)? @bar)* @foos
    (bar (_)? @baz)+ @bars

The quantifier of `@bar` would be `*` (combining `*` and `?`), and the value a list. The quantifier for `@baz` would also be `*` (combining `+` and `?`).

If logic is necessary to tie one match to another, for example an element of `@foos` with the corresponding element in `@bar`, this can always be expressed by writing a different stanza that matches `(foor (_)? @bar) @foo`.

hendrikvanantwerpen:
After some discussion with @dcreager we decided to start with a conservative approach, and see how that works. That means the general `if` and `for` are for now put on hold, although for now we keep the [proof of concept](https://github.com/tree-sitter/tree-sitter-graph/tree/poc/lazy-evaluation-of-loops-and-conditionals) that shows how it can be implemented with lazy evaluation.

Instead, we opt for an approach where the control flow is directly controlled by the AST and the results of matches, but not by any other computed values. For `if` and `for` that means that the values must be syntactic captures, as in `for x in @xs` and `if some @x` or `if none @x`. The same restriction has been applied to `scan`, which can only scan captures, global variables, or literal strings (which is convenient for testing).

The result of this is that the lazy evaluation that is necessary for these constructs is much simpler than if they took arbitrary values as input. All of `for`, `if`, and `scan` can be unfolded before any lazy values need to be forced. This means the lazy evaluation code has no concept of these constructs and of iteration anymore, resulting in cleaner code and smaller data structures.

To make reviewing and merging a bit more controlled, I have split up the original large PR that this was into a string of smaller ones.

tausbn:
> After some discussion with @dcreager we decided to start with a conservative approach, and see how that works. That means the general `if` and `for` are for now put on hold, although for now we keep the [proof of concept](https://github.com/tree-sitter/tree-sitter-graph/tree/poc/lazy-evaluation-of-loops-and-conditionals) that shows how it can be implemented with lazy evaluation.

I think that's a good idea. As I've mentioned elsewhere, it feels like this control-flow should only exist at the outer layer of the stanzas (or rather, as multiple potential "disjuncts" for each stanza). This would move the potential branching all the way up to the level of the query itself.

> Instead, we opt for an approach where the control flow is directly controlled by the AST and the results of matches, but not by any other computed values. For `if` and `for` that means that the values must be syntactic captures, as in `for x in @xs` and `if some @x` or `if none @x`. The same restriction has been applied to `scan`, which can only scan captures, global variables, or literal strings (which is convenient for testing).

I assume "only scan captures" includes scanning of things _derived_ from captures (e.g. `(source-text @foo)`)? Otherwise I'm not sure what you can usefully get from scanning a syntax node (and more to the point, the types wouldn't match up).

Also, would there be any way of checking for the emptiness of a list-like capture? (Say I want to do some setup before doing `for x in @xs`, but that this setup should only be run once.)

Finally, my understanding of the proposed semantics is that 
```
(foo (bar)? @bar)+ @foos
```
would result in two lists, one with `foo` nodes, and one with (potentially) `bar` nodes. However, I'm struggling to see how to actually _use_ such a setup. Surely, if I want to reason based on a specific `foo` and `bar` node (assuming both are present), I need to know the link between them, and producing them as lists is not conducive to this.

One solution would be to produce an explicit `none` for the cases where the `bar` node is not present. If we then have a suitable `zip` function, we can traverse both of these lists in tandem, and check whether the `bar` component is `none` or not. This would require an extension to the `for` syntax, though (or some other way of decomposing the tuple).

There's also something oddly asymmetrical about this setup that I can't quite put my finger on. The fact that I could also write the above query as 
```
((foo (bar)? @bar) @foos)+
```
(thus making `@foos` a list by virtue of its inherited list-like container) makes me a bit uneasy, as it suggests we're doing something overly specific.

One idea would be to say that the result of `(...)+ @foos` is a list of tuples, with each entry coming from the captures inside the `...`. Essentially, the same list of tuples as produced by the zipping operation I mention above. If the `...` doesn't contain additional captures, then it's just a list of the syntax nodes that matched this bit.

The only slightly awkward bit about this suggestion is that currently all captures must be named (and so will result in corresponding lists for the inner captures, which is wasteful). One solution to this awkwardness would be to introduce an anonymous capture, call it `@_`.

Then, 
```
(foo (bar)? @_)+ @bars
```
would capture just the `bar` nodes that were present,
```
((foo (bar)? @_) @_)+ @bar_foos
```
would capture a list of `(bar node or none, foo node)` tuples (corresponding to the order of the anonymous captures),
```
((foo (bar)? @bars) @foos)+ 
```
would capture two lists, one with `foo` nodes and one with `bar` nodes (and no `none`s), and finally
```
((foo (bar)? @bars) @foos)+ @bar_foos
```
would give you all of the above, simultaneously.

One thing I don't like about this, though, is that it would put the `bar` nodes before the `foo` nodes in each tuple, and this seems counterintuitive. (Actually, there are many things I don't like about this, but I've written it now, and maybe it'll help spark some better ideas. :slightly_smiling_face:)

hendrikvanantwerpen:
> I assume "only scan captures" includes scanning of things derived from captures (e.g. (source-text @foo))? Otherwise I'm not sure what you can usefully get from scanning a syntax node (and more to the point, the types wouldn't match up).

Hm, that is a good point. I was assuming that `scan @id` is allowed, but that might be based on the old DSL. If an explicit `source-text` is necessary, that would be too limiting.

I suppose in that case I have to be a little more sophisticated about checking which terms are allowed, and not only rely on top-level syntax forms. But, this is not too difficult to achieve, I believe.

hendrikvanantwerpen:
> Also, would there be any way of checking for the emptiness of a list-like capture? (Say I want to do some setup before doing for x in @xs, but that this setup should only be run once.)

That would be possible without changing the fundamental way we evaluate things. Initially I thought about introducing some special forms (like `if empty @xs`). But, if `scan` requires a kind of `const` analysis to be useful, perhaps having a conditional that only allows const expressions like `if (list-empty @xs)` might be better. I'll think about this!

hendrikvanantwerpen:
> Finally, my understanding of the proposed semantics is that `(foo (bar)? @bar)+ @foos` would result in two lists, one with foo nodes, and one with (potentially) bar nodes. However, I'm struggling to see how to actually use such a setup. Surely, if I want to reason based on a specific foo and bar node (assuming both are present), I need to know the link between them, and producing them as lists is not conducive to this.

I think your understanding of the semantics is correct. A lot of the structure of the match is lost, already when we get the results back from tree-sitter. As a practical approach to solve this, I think you can write a separate stanza `(foo (bar)? @bar) @foo` where you handle the specifics of the pair, and use the stanza over the list to do the list specific behavior.

FoamScience:
Hi @tausbn @hendrikvanantwerpen , this is truly incredible! However I have stumbled upon some things which I think are not intended.
As a simple example, the following will try to mark abstract classes in a C++ file (by hunting for pure virtual member methods):
```scheme
;; Create nodes for classes
(class_specifier
    name: (_) @class
)
{
    node @class.node
    ;; Approach 01: Using a scoped variable
    let @class.ab = ""      ;; class assumed to be concrete
}

(class_specifier
    name: (_) @class
    (field_declaration_list
        (field_declaration
            (virtual_function_specifier)? @virtual
            type: (_) @type
            declarator: (function_declarator) @member
            default_value: (number_literal)? @pure
        )
    )
)
{
    if some @pure, some @virtual {
        ;; The following line will mysteriously say "Duplicate @class.ab"; does the if statement handle scoped vars as intended?
        set @class.ab = "true"
        ;; Approach 02: Using graph node attributes
        ;; This will say "Duplicate attribute"; is there a way to check if an attribute is already set? maybe I missed something
        ;;attr (@class.node) abstract = "true"
    }
}

```

dcreager (approved):
This is great!

PR #25, opened 2021-12-22
Fix nested scan

hendrikvanantwerpen:
The scan statement reuses a vector to store regex captures. This is incorrect if scans are nested, because the inner scan overwrites the captures of the outer scan, even though these might be used in the next iteration of the outer scan. Instead of reusing the vector, a new vector is used, preserving any enclosing captures.

### DIFF
--- a/src/reference/mod.rs
+++ b/src/reference/mod.rs
@@ -357,6 +357,8 @@
 //! earliest matching regular expression in each iteration, until we have exhausted the entire
 //! string, or none of the regular expressions match.
 //!
+//! Note that a regular expression that matches an empty string results in an error.
+//!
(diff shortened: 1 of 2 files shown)


src/reference/mod.rs
```diff
@@ -357,6 +357,8 @@
 //! earliest matching regular expression in each iteration, until we have exhausted the entire
 //! string, or none of the regular expressions match.
 //!
+//! Note that a regular expression that matches an empty string results in an error.
```
dcreager, line 360: Oh right, we need this constraint for a termination guarantee, don't we?

dcreager (approved):

src/reference/mod.rs
hendrikvanantwerpen: Indeed! Found this out the hard way...

PR #26, opened 2021-12-22
Implement block scoping

hendrikvanantwerpen:
Variable environments where reused inside blocks for scan arms. This is incorrect, because (a) later ietartions of a block might cause duplicate variable errors for local variables, and (b) variables defined inside the block persist after it. This is fixed by creating a hierarchy of environments, where environments delegate to their parent for unknown variables.

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -5,7 +5,7 @@
 // Please see the LICENSE-APACHE or LICENSE-MIT files in this distribution for license details.
 // ------------------------------------------------------------------------------------------------
 
-use std::collections::HashMap;
+mod variables;
 
 use anyhow::Context as ErrorContext;
 use thiserror::Error;
@@ -36,15 +36,17 @@ use crate::ast::Statement;
 use crate::ast::StringConstant;
 use crate::ast::UnscopedVariable;
 use crate::ast::Variable;
+pub use crate::execution::variables::Globals;
+use crate::execution::variables::ScopedVariables;
+use crate::execution::variables::VariableMap;
+use crate::execution::variables::Variables;
 use crate::functions::Functions;
 use crate::graph::DisplayWithGraph;
 use crate::graph::Graph;
 use crate::graph::GraphNodeRef;
-use crate::graph::SyntaxNodeRef;
 use crate::graph::Value;
 use crate::Context;
 use crate::DisplayWithContext;
-use crate::Identifier;
 
 impl File {
     /// Executes this graph DSL file against a source file.  You must provide the parsed syntax
@@ -56,13 +58,13 @@ impl File {
         tree: &'tree Tree,
         source: &'tree str,
         functions: &mut Functions,
-        globals: &mut Variables,
+        globals: &Globals,
     ) -> Result<Graph<'tree>, ExecutionError> {
         let mut graph = Graph::new();
         if tree.root_node().has_error() {
             return Err(ExecutionError::ParseTreeHasErrors);
         }
-        let mut locals = Variables::new();
+        let mut locals = VariableMap::new();
         let mut scoped = ScopedVariables::new();
         let current_regex_captures = Vec::new();
         let mut function_parameters = Vec::new();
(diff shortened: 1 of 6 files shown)


src/execution.rs
```diff
@@ -5,14 +5,17 @@
...
 
 use anyhow::Context as ErrorContext;
 use thiserror::Error;
 use tree_sitter::QueryCursor;
 use tree_sitter::QueryMatch;
 use tree_sitter::Tree;
 
+pub use variables::Globals;
+use variables::*;
+
```
dcreager, line 18: nit: Can you mimic the `use` style down below?  Prefix these with `crate::`, add them to the section below (`rustfmt` will auto-sort them alphabetically for you), and list out the specific types that you're using instead of using a `*` import.

src/execution.rs
```diff
@@ -5,14 +5,17 @@
 // Please see the LICENSE-APACHE or LICENSE-MIT files in this distribution for license details.
 // ------------------------------------------------------------------------------------------------
 
-use std::collections::HashMap;
+mod variables;
```
dcreager, line 8: There are a lot of `pub` types in variables.rs, which means that either this needs to be `pub mod` (so that external callers can name the types), or the types in variables.rs should actually be `pub(crate)`.

How do you feel about either of the following to clean it up a bit:

- Make `variables` a proper top-level module (so a sibling file to execution.rs, with a `pub mod` statement in lib.rs).  Change `pub use` below into `use`, since callers will name the type `tree_sitter_graph::variables::Globals` instead of `tree_sitter_graph::execution::Globals`.

- If most of the content of variables.rs is really local to execution.rs, that tells me it doesn't deserve to be a separate file, and all of the variable-related should still live here in execution.rs.  You can use section comments like [this](https://github.com/github/stack-graphs/blob/e17b7fc0d9555de333d4b27a7efe95e84c50401d/stack-graphs/src/graph.rs#L68-L69) to organize the contents of the file more clearly.

hendrikvanantwerpen: Good catch. I wasn't entirely sure how the visibility should work here.

Let me explain what I wanted to achieve first:

- Most of `variables` is internal to `execution`.
- I personally don't like working with large files. I find `execution` already too large as it is, but it makes sense that these things are together. That's why I wanted to factor out the `variables` related types into that separate module.
- `Globals` is a something that is part of the crate interface. It uses `VariableMap` internally, so it seemed convenient to define it in the same module.
- I thought the fact that `variables` is `mod variables` in `execution` essentially meant it was only visible inside `execution`.

In light of this, I'd rather not merge `variables` into `execution` (and for `lazy_execution` there would be even of that, making that file even larger), but I don't think it should be a top-level module either. Scoping it as `pub(crate)` would be possible, but the scope could be even smaller---if that can be expressed.

Also, I don't think I understand how `mod`/`pub(X) mod` of the module interacts with the `pub(X)` definitions inside the module.

hendrikvanantwerpen: After reading up on Rust modules, I think my idea is achieved with `pub(super)` visibility. I've updated the code to use that wherever possible. Is this an acceptable solution, or do you still have concerns?

dcreager: I'm okay with that!

src/execution.rs
```diff
@@ -141,111 +143,30 @@ impl ExecutionError {
     }
 }
 
-struct ExecutionContext<'a, 'tree> {
+struct ExecutionContext<'a, 'g, 's, 'tree> {
```
dcreager, line 146: One last nit: Do you know if this works without the extra lifetime parameters?  i.e., using `'a` down below as the parameters to `Globals` and `ScopedVariables`.  Often the compiler will be able to infer a "containing" lifetime, and that will simplify the types here a bit.

dcreager (approved):
[forgot to approve, I'm :+1: modulo the question about lifetime parameters]

src/execution.rs
hendrikvanantwerpen: I tried that first, but could not get it to work everywhere. I'll keep it like this, but if you think it is possible, I'm open to pairing on this to figure it out.

PR #28, opened 2021-12-22
Update builtin functions

hendrikvanantwerpen:
Renames  to the correct , adds several general operations, and groups related functions in modules in the source.


tausbn (approved):
This looks good to me. There are a bunch more convenience functions I could think of adding, but most of these can wait until they're needed.

dcreager (approved):

PR #31, opened 2022-01-06
Add execute_into method

dcreager:
This variant allows you to execute a TSG file into an _existing_ `Graph` instance, which is useful if you need to create predefined nodes and/or edges before executing.

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -61,6 +61,24 @@ impl File {
         globals: &Globals,
     ) -> Result<Graph<'tree>, ExecutionError> {
         let mut graph = Graph::new();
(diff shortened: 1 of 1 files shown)


hendrikvanantwerpen (approved):

hendrikvanantwerpen:
This is pretty useful when trying to get CLI + stack graphs working. Perhaps it is better to do this patch against `main` while we hold of on the context removal?

dcreager:
> This is pretty useful when trying to get CLI + stack graphs working. Perhaps it is better to do this patch against `main` while we hold of on the context removal?

Done!  I reordered this PR and #13 so that we can review / merge this first.

PR #32, opened 2022-01-17
Add value conversions

hendrikvanantwerpen:
This adds:
- two owned `into_*` value conversions for graph and syntax references, and
- borrowed `as_*` conversions as an alternative to the owned conversions, which sometimes force unnecessary clones.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -459,17 +459,27 @@ impl Value {
 
     /// Coerces this value into a syntax node reference, returning an error if it's some other type
     /// of value.
-    pub fn into_syntax_node<'a, 'tree>(
+    pub fn into_syntax_node_ref<'a, 'tree>(
         self,
         graph: &'a Graph<'tree>,
-    ) -> Result<&'a Node<'tree>, ExecutionError> {
+    ) -> Result<SyntaxNodeRef, ExecutionError> {
         match self {
-            Value::SyntaxNode(node) => Ok(&graph[node]),
+            Value::SyntaxNode(node) => Ok(node),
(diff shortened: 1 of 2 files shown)


src/graph.rs
```diff
@@ -424,6 +424,16 @@ impl Value {
         }
     }
 
+    pub fn as_bool(&self, graph: &Graph) -> Result<&bool, ExecutionError> {
```
dcreager, line 427: For `as_bool` and `as_integer`, I think it's worth having the return value be `bool` / `u32` instead of `&bool` / `&u32`.

src/graph.rs
```diff
@@ -446,6 +466,16 @@ impl Value {
         }
     }
 
+    pub fn as_string(&self, graph: &Graph) -> Result<&String, ExecutionError> {
```
dcreager, line 469: And I think it's more idiomatic for the return type here to be `&str` instead of `&String`

src/graph.rs
```diff
@@ -457,14 +487,75 @@ impl Value {
...
+                "got {}",
+                self.display_with(graph)
+            ))),
+        }
+    }
+
+    pub fn as_graph_node_ref<'a, 'tree>(
+        &self,
+        graph: &'a Graph<'tree>,
+    ) -> Result<&GraphNodeRef, ExecutionError> {
```
dcreager, line 518: Same as above for `as_{graph,syntax}_node_ref` — `{Graph,Syntax}NodeRef` are just `u32`s under the covers, so I think we should return them by value not by reference

dcreager (approved):
:+1: modulo the suggestions below

PR #33, opened 2022-01-17
Lazy evaluation

hendrikvanantwerpen:
Adds lazy evaluation as an alternative execution strategy.

**Why**

Strict evaluation required careful ordering of the stanzas to make sure scoped variables were defined before they were used. This can get in the way of writing readable code.

**How**

The lazy evaluation strategy executes all stanzas on a given tree, but does not compute the values immediately. Instead, it computes thunks, that are forced only after _all_ stanzas have been executed when the graph is built in the second phase. The key part are the handling of scoped variables: initially both scope and value are thunks. Whenever a scoped variable is forced, all scopes for that scoped variable name are forced, but not yet their values, which are only forced when the scoped variable for its particular scope is forced.

**Limitations**

The lazy values are relatively simple and do not support control. This means control statements cannot depend on thunks, but only on values that are directly tied to the tree. While supporting lazy control is possible (see this [proof of concept](https://github.com/tree-sitter/tree-sitter-graph/tree/poc/lazy-evaluation-of-loops-and-conditionals)), it is not clear that the added complexity and possible runtime cost is justified.

**Use**

The API has new methods `execute_lazy{,_into}` that have the same signature as the (still existing) `execute{,_into}` methods, but use the new lazy evaluation strategy.

The CLI has a new `--lazy/-z` option to use the lazy strategy. The default is still the current, strict evaluation.

### DIFF
--- a/src/bin/tree-sitter-graph/main.rs
+++ b/src/bin/tree-sitter-graph/main.rs
@@ -35,6 +35,12 @@ fn main() -> Result<()> {
                 .long("quiet")
                 .help("Suppress console output"),
         )
+        .arg(
+            Arg::with_name("lazy")
+                .short("z")
+                .long("lazy")
+                .help("Use lazy evaluation (experimental)"),
+        )
         .arg(Arg::with_name("scope").long("scope").takes_value(true))
         .arg(Arg::with_name("json").long("json").takes_value(false))
         .get_matches();
@@ -43,6 +49,7 @@ fn main() -> Result<()> {
     let source_path = Path::new(matches.value_of("source").unwrap());
     let current_dir = std::env::current_dir().unwrap();
     let quiet = matches.is_present("quiet");
+    let lazy = matches.is_present("lazy");
     let config = Config::load()?;
     let mut loader = Loader::new()?;
     let loader_config = config.get()?;
@@ -65,9 +72,12 @@ fn main() -> Result<()> {
         .with_context(|| anyhow!("Error parsing TSG file {}", tsg_path.display()))?;
     let mut functions = Functions::stdlib(&mut ctx);
     let globals = Variables::new();
-    let graph = file
-        .execute(&ctx, &tree, &source, &mut functions, &globals)
-        .with_context(|| format!("Could not execute TSG file {}", tsg_path.display()))?;
+    let graph = if lazy {
+        file.execute_lazy(&mut ctx, &tree, &source, &mut functions, &globals)
+    } else {
+        file.execute(&mut ctx, &tree, &source, &mut functions, &globals)
+    }
+    .with_context(|| anyhow!("Could not execute TSG file {}", tsg_path.display()))?;
     let json = matches.is_present("json");
     if json {
         graph.display_json(&ctx);
--- a/src/bin/tree-sitter-graph/main.rs
+++ b/src/bin/tree-sitter-graph/main.rs
@@ -23,6 +23,8 @@ use tree_sitter_loader::Loader;
 const BUILD_VERSION: &'static str = env!("CARGO_PKG_VERSION");
 
 fn main() -> Result<()> {
+    init_log();
+
     let matches = App::new("tree-sitter-graph")
         .version(BUILD_VERSION)
         .author("Douglas Creager <dcreager@dcreager.net>")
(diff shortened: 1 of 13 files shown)


src/lazy_execution/values.rs
```diff
@@ -0,0 +1,354 @@
...
+use crate::DisplayWithContext;
+use crate::Identifier;
+
+use super::store::*;
+use super::DisplayWithContextAndGraph;
+use super::EvaluationContext;
+
+/// Lazy values
+#[derive(Clone, Debug)]
+pub enum Value {
```
dcreager, line 31: This might seem like a cumbersome suggestion, but can we rename this to `LazyValue` (ditto the other types in this module)?  I actually didn't realize until this point in the PR that all of the references to `Value` earlier on were referring to a new type.

Unless the intent is that these types will _replace_ the equivalently named ones in the `graph` module once we remove the old evaluation strategy?  If that's the case, then we can deal with a bit of confusion in the meantime.  But if both will exist even once lazy evaluation is the only strategy that's implemented, then I think it's important for the names to clarify the difference between a lazy value and a forced strict value.

src/lazy_execution/variables.rs
```diff
@@ -0,0 +1,166 @@
...
+use crate::DisplayWithContext;
+use crate::Identifier;
+
+use super::store::DebugInfo;
+use super::store::Store;
+use super::values::Value;
+use super::DisplayWithContextAndGraph;
+
+/// An environment of named lazy variables
+pub trait Variables {
```
dcreager, line 23: This, for instance, _wouldn't_ need to be renamed to `LazyVariables`, because I anticipate that the other non-lazy `Variables` trait would go away when we remove the old evaluation strategy.

tests/it/lazy_execution.rs
```diff
@@ -0,0 +1,1255 @@
...
+}
+
+fn check_execution(python_source: &str, dsl_source: &str, expected_graph: &str) {
+    match execute(python_source, dsl_source) {
+        Ok(actual_graph) => assert_eq!(actual_graph, expected_graph),
+        Err(e) => panic!("Could not execute file: {}", e),
+    }
+}
+
+#[allow(unused)]
```
dcreager, line 50: You can remove this annotation since the function is used in some of the tests below

dcreager (commented):
This looks great!  Some initial comments:

src/lazy_execution/store.rs
```diff
@@ -0,0 +1,296 @@
...
+            None => {
+                return Err(ExecutionError::UndefinedScopedVariable(format!(
+                    "{}.{}",
+                    scope.display_with(exec.graph),
+                    name.display_with(exec.ctx),
+                )));
+            }
+        };
+        match values.replace(ScopedValues::Forcing) {
+            ScopedValues::Unforced(pairs) => {
```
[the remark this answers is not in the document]
hendrikvanantwerpen: > My hunch is that with this pattern it was easier to make the borrow checker happy?

Exactly! I wasn't too happy with it either, but I couldn't get anything else to work.

src/bin/tree-sitter-graph/main.rs
```diff
@@ -35,6 +37,12 @@ fn main() -> Result<()> {
                 .long("quiet")
                 .help("Suppress console output"),
         )
+        .arg(
+            Arg::with_name("lazy")
```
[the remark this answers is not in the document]
hendrikvanantwerpen: Indeed

src/lazy_execution/store.rs
```diff
@@ -0,0 +1,296 @@
...
+                debug_info
+            )
+        })?;
+        Ok(value)
+    }
+}
+
+/// Data structure to hold scoped variables with lazy keys and values
+pub struct ScopedVariables {
+    variables: HashMap<Identifier, Rc<RefCell<ScopedValues>>>,
```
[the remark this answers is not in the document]
hendrikvanantwerpen: That works!

src/lazy_execution/variables.rs
hendrikvanantwerpen: Indeed. I've already merged them for another PR I'm working on, actually.

src/lazy_execution/values.rs
hendrikvanantwerpen: > Unless the intent is that these types will replace the equivalently named ones in the graph module once we remove the old evaluation strategy?

No, they'll always be separate. I can prefix them to make the code more clear.

dcreager (approved):
:tada:

PR #36, opened 2022-02-03
Allow all local expression in control statements

hendrikvanantwerpen:
Currently control statements restrict their argument to certain syntactic classes. This is overly restrictive, as it disallows `scan (source-text @capture)`, for example.

This PR introduces a checker that is executed after the parser, that analyzes whether expressions are local (i.e., only depend on the stanza match and not on scoped variables), and allows all those expressions as argument to the control statement. Additionally, the checker computes capture indices say we don't have to do lookups by name when executing.

Possible disadvantage:
- Slower parsing times because an extra pass over the AST is made.

Possible advantages:
- Faster execution (especially if the same `File` instance is applied multiple times), because name lookups for captures are done only once.
- Cleaner separation of parsing and semantic checks. This also makes it easier to swap the parser for something else (e.g., tree-sitter based), if we want to do that in the future.


src/checker.rs
```diff
@@ -0,0 +1,544 @@
...
+        if !value_result.is_local {
+            return Err(CheckError::ExpectedLocalValue(self.location));
+        }
+
+        for arm in &mut self.arms {
+            // Be aware that this check is not complete, as it does not rule out
+            // all regular expressions that admit empty matches. For example, th
+            // regex "\b" matches empty strings within a larger non-empty one.
+            // Therefore, there is also a runtime check that checks that a match was
+            // non-empty. This is all to prevent non-termination of scan.
```
dcreager, line 190: Damn you, Turing!

src/checker.rs
```diff
@@ -0,0 +1,544 @@
...
+        &mut self,
+        ctx: &mut CheckContext,
+        value: ExpressionResult,
+        mutable: bool,
+    ) -> Result<(), CheckError> {
+        let mut value = value;
+        // Mutable variables are not considered local, because a non-local
+        // assignment in a loop could invalidate an earlier local assignment.
+        // Since we process all statement in order, we don't have info on later
+        // assignments, and and assume non-local to be sound.
```
dcreager, line 468: nit: "and and" ⇒ "and can"?

src/checker.rs
```diff
@@ -0,0 +1,544 @@
...
+    fn set_check(
+        &mut self,
+        ctx: &mut CheckContext,
+        value: ExpressionResult,
+    ) -> Result<(), CheckError> {
+        let mut value = value;
+        // Mutable variables are not considered local, because a non-local
+        // assignment in a loop could invalidate an earlier local assignment.
+        // Since we process all statement in order, we don't have info on later
+        // assignments, and and assume non-local to be sound.
```
dcreager, line 486: ditto

src/parser.rs
```diff
@@ -23,7 +23,23 @@ use crate::Identifier;
...
+        #[allow(deprecated)]
+        file.parse(ctx, source)?;
+        file.check(ctx)?;
+        Ok(file)
+    }
+
     /// Parses a graph DSL file, adding its content to an existing `File` instance.
+    #[deprecated(
+        note = "Parsing multiple times into the same `File` instance is unsound. Use `File::from_source` instead."
+    )]
```
dcreager, line 42: Interesting!  I had originally added this method in the hopes that we'd implement "include files" or something similar using it.  (Hand waving: you'd pass multiple `.tsg` files on the command line, and we'd call `parse` on each one in order to build up the combined TSG file before executing.)

There's definitely no need to tie ourselves in knots supporting that pattern, especially since we haven't actually implemented include files yet.  Whenever we do, we can change things around as needed at the API level.

But, the deprecation comment makes we wonder whether the unsoundness that you're describing will _preclude_ the possibility of include files?

dcreager (approved):

src/parser.rs
hendrikvanantwerpen: > I had originally added this method in the hopes that we'd implement "include files" or something similar using it.

Interesting idea! I hadn't thought of that use case.

I think this method was perfectly fine until I implemented query merging. Because I merge all queries into one at the end of a parse, it would overwrite an earlier merged query. If we want to support this, we must make sure we merge all queries from all files correctly. (If only we could incrementally add patterns to query...) Another question is if the computed capture indices in previous files remain valid in the new merged query. This might all work out nicely, or be annoying, but I'm saving that work for when actually looking at imports :)

src/checker.rs
hendrikvanantwerpen: Ain't he the sneaky fellow...

PR #41, opened 2022-02-10
Paper cuts

hendrikvanantwerpen:


src/variables.rs
```diff
@@ -125,6 +125,10 @@ impl<'a> Globals<'a> {
         Self(VariableMap::new())
     }
 
+    pub fn new_child(parent: &'a mut Globals<'a>) -> Self {
```
tausbn, line 128: The word "child" here threw me for a bit, but I see now that this is not merely a map from variables to values, but a _tree_ of such maps. I feel like this ought to be documented more clearly, but I guess that doesn't have to be in this PR.

src/functions.rs
```diff
@@ -128,6 +128,8 @@ impl Functions {
         functions.add(Identifier::from("plus"), stdlib::math::Plus);
         // string functions
         functions.add(Identifier::from("replace"), stdlib::string::Replace);
+        // list functions
+        functions.add(Identifier::from("is-empty"), stdlib::list::IsEmpty);
```
tausbn, line 132: Presumably there should also be a call that adds the `length` function?

tausbn (changes requested):
A few comments, but otherwise this looks good to me. :slightly_smiling_face:

src/variables.rs
hendrikvanantwerpen: Good point! I can definitely add a comment. If you have a suggestion for a better name, I'm also open to that.

src/functions.rs
hendrikvanantwerpen: Yes, there should!

src/variables.rs
hendrikvanantwerpen: For example, `new_nested` or `new_inherit` (or `new_inheriting`?).

tausbn: Hmm... `new_environment`, perhaps? Though I do like the explicitness of saying it's nested/inherited. One could also use the word "scope", though I feel like that's potentially overloaded already in this context.

tausbn (approved):

PR #43, opened 2022-02-16
Add explicit global variable declarations

hendrikvanantwerpen:
Adds explicit global variable declarations to the DSL. For example:

    global filename

    (module) {
      scan filename {
        ; ...
      }
    }

These declarations are **required** for any global variable that is used in stanzas.

This has the following advantages:
- DSL files are now completely self-contained (no free variables in stanzas), and explicit about their requirements for execution.
- Variable references can now be checked at parse time, and errors are caught without having to hit a stanza that uses the missing variable.
- When executing a DSL file, the global environment is checked if all required variables are there. Again, giving early feedback of problems, instead of having to execute until the missing variable is used.

Using undeclared globals is deprecated with this change. A warning is displayed when an undefined variable is encountered (which is then assumed to be global) that it should be declared. Eventually, declaring all globals will be required.

See this example for how they can be used: https://github.com/hendrikvanantwerpen/tree-sitter-typescript/blob/add-stack-graph/queries/stack-graphs.tsg#L13-L17

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -22,6 +22,12 @@ use crate::Location;
 
 #[derive(Debug, Error)]
 pub enum CheckError {
+    #[error("Cannot hide global variable {0} at {1}")]
+    CannotHideGlobalVariable(String, Location),
+    #[error("Cannot set global variable {0} at {1}")]
+    CannotSetGlobalVariable(String, Location),
+    #[error("Duplicate global variable {0} at {1}")]
+    DuplicateGlobalVariable(String, Location),
     #[error("Expected list value at {0}")]
     ExpectedListValue(Location),
     #[error("Expected local value at {0}")]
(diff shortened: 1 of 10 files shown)


src/checker.rs
```diff
@@ -479,16 +527,21 @@ impl ast::UnscopedVariable {
...
-            .get(&self.name)
-            .cloned()
-            .unwrap_or_else(|| ExpressionResult {
+        if let Some(result) = ctx.globals.get(&self.name) {
+            Some(result)
+        } else {
+            ctx.locals.get(&self.name)
+        }
+        .cloned()
+        // TODO deprecated support for undeclared globals
```
dcreager, line 536: Do we need to keep this?  @robrix @tausbn, I think you're the primary current users of this, would it be problematic to go ahead and make undeclared globals a hard error now?

dcreager (approved):
This is a _great_ quality-of-life addition!

Can you add a description of this feature to the changelog, creating an "unreleased" section at the top if there isn't already one?  That will make it easier to create the correct changelog when we eventually cut a release with this new feature.

hendrikvanantwerpen:
@dcreager Good point about the changelog! Will do.

PR #44, opened 2022-02-17
relax parse error behavior

hendrikvanantwerpen:
- Improve display of parse errors
- Move parse error check to CLI
- Add flag to allow parse errors
- Skip tree traversal unless there are errors

Example verbose error:

```
Unexpected syntax on line 3 column 17:

| type U = { X: { f number } };
                  ^

```

Example compact error:

```
Unexpected syntax on line 3 column 17: f
```

_Can be reviewed commit-by-commit_


src/bin/tree-sitter-graph/util.rs
```diff
@@ -0,0 +1,84 @@
...
+        } else {
+            None
+        };
+        self.errors.push((start.row, start.column, node_source));
+    }
+
+    pub(crate) fn is_empty(&self) -> bool {
+        self.errors.is_empty()
+    }
+}
```
tausbn, line 65: It would be nice to have a way to access the errors, beyond just displaying them or checking that there are none. 

The way I currently handle errors is to recognise them in a stanza, which is a bit awkward (since you can't actually query for `ERROR` nodes). However, it occurs to me that if I could access the list of errors, then I could just produce the relevant graph (consisting of a single "syntax error" node) directly, and do away with needlessly running the `tree-sitter-graph` machinery. (And in fact, I would be happy with just the _first_ syntax error in the file.)

src/bin/tree-sitter-graph/util.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    fn add(&mut self, node: tree_sitter::Node, source: &str) {
+        let start = node.start_position();
+        let node_source = if !node.byte_range().is_empty() {
+            Some(source[node.byte_range()].to_string())
+        } else {
+            None
+        };
+        self.errors.push((start.row, start.column, node_source));
```
tausbn, line 59: For further handling, it would be nice to have also the end row/column available. 
(For displaying the error, however, I don't think this is necessary.)

src/bin/tree-sitter-graph/main.rs
```diff
@@ -67,6 +76,17 @@ fn main() -> Result<()> {
     let tree = parser
         .parse(&source, None)
         .ok_or_else(|| anyhow!("Could not parse {}", source_path.display()))?;
+    let allow_parse_errors = matches.is_present("allow-parse-errors");
+    if !allow_parse_errors {
+        let parse_errors = ParseErrors::from_tree(&tree, &source);
```
tausbn, line 81: I wonder if it would be a bit more smooth to have `from_tree` return an `Option<ParseError>` (that could then be decomposed in an `if let`). That way, you wouldn't need `is_empty` at all. 

Though of course then you have the awkwardness of `Some(vec![])` being a potential value that should never appear.

Hm... Maybe it would suffice to just display the _first_ syntax error in the file? (Which I believe will be the first error node encountered during the tree walk.) After all, after the first syntax error, the rest of the parse could be completely out of whack.

tausbn (dismissed):
I have a few comments and suggestions, but apart from that this looks really nice! I like the way you've moved the check for syntax errors into the main function. That'll make it a lot easier to set up custom handling of syntax errors for my purposes. :+1:

hendrikvanantwerpen:
If I understand you correctly, you would also want to use this machinery yourself for finding/reporting errors.

This PR does not actually add the parse error logic to the API, it's only internal. I see that it could be useful for others as well, although that requires a bit more thought---along the lines of your comments. I think such logic should perhaps become part of the tree-sitter API, instead off this one.

On a practical note, I'm off for the next two weeks, so i won't be able to think about these things for a while. I think this PR is quite safe, since it does not change the API, so no external code that we may break later.

hendrikvanantwerpen:
@tausbn I have made some changes so that the error finding and display is now part of the lib instead of the bin, so it should be reusable. There are methods for finding all or just the first error, and error display has a flag that controls whether you get nice multi-line errors, or more compact single line errors.

src/util.rs
```diff
@@ -0,0 +1,140 @@
+// -*- coding: utf-8 -*-
```
dcreager, line 1: I would suggest a different filename — `util` is very generic.  Maybe `parse_error.rs`?

dcreager (approved):
Very nice.  Not in scope for this PR, but I've seen recommendations for the [`miette`](https://docs.rs/miette/*/miette/) if in the future we want to make the verbose error output more detailed and rustc-like.

src/util.rs
hendrikvanantwerpen: That is a good idea.

PR #48, opened 2022-03-17
Add support for attribute shorthands

hendrikvanantwerpen:
This adds support for shorthand attributes, attributes that are expanded to a set of other attributes. For example:

``` tsg
attribute node_props = node => node_text = (source-text node), node_index = (named-child-index node)

(argument (_)@expr) {
  attr (@expr) node_props = @expr
}
```

See this example for how they can be used to introduce compact notation for common attribute combinations: https://github.com/hendrikvanantwerpen/tree-sitter-typescript/blob/add-stack-graph/queries/stack-graphs.tsg#L19-L35

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -752,13 +752,11 @@ impl AttributeShorthands {
         self.0.insert(shorthand.name.clone(), shorthand);
     }
 
-    pub fn iter(&self) -> std::collections::hash_map::Values<Identifier, AttributeShorthand> {
+    pub fn iter(&self) -> impl Iterator<Item = &AttributeShorthand> {
         self.0.values()
     }
 
-    pub fn into_iter(
-        self,
-    ) -> std::collections::hash_map::IntoValues<Identifier, AttributeShorthand> {
+    pub fn into_iter(self) -> impl Iterator<Item = AttributeShorthand> {
         self.0.into_values()
     }
 }
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -8,6 +8,7 @@
 //! Defines the AST structure of a graph DSL file
 
 use regex::Regex;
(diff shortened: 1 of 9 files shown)


src/ast.rs
```diff
@@ -730,3 +734,50 @@ impl From<ScopedVariable> for Expression {
...
+
+    pub fn get(&self, name: &Identifier) -> Option<&AttributeShorthand> {
+        self.0.get(name)
+    }
+
+    pub fn add(&mut self, shorthand: AttributeShorthand) {
+        self.0.insert(shorthand.name.clone(), shorthand);
+    }
+
+    pub fn iter(&self) -> std::collections::hash_map::Values<Identifier, AttributeShorthand> {
```
dcreager, line 755: Can you make the return type opaque?  I think it would be something like

``` rust
pub fn iter(&self) -> impl Iterator<Item=&AttributeShorthand> + '_
```
or maybe

``` rust
pub fn iter(&self) -> impl Iterator<Item=&'_ AttributeShorthand> + '_
```

src/ast.rs
```diff
@@ -730,3 +734,50 @@ impl From<ScopedVariable> for Expression {
...
+        self.0.values()
+    }
+
+    pub fn into_iter(
+        self,
+    ) -> std::collections::hash_map::IntoValues<Identifier, AttributeShorthand> {
```
dcreager, line 761: Same here:

``` rust
pub fn into_iter(self) -> impl Iterator<Item=AttributeShorthand>
```

dcreager (approved):
Just a couple of nits

src/ast.rs
hendrikvanantwerpen: Brilliant! I didn't know how to do this.

PR #49, opened 2022-03-17
Implicit debug attrs

hendrikvanantwerpen:
This adds an option to have tree-sitter-graph implicitly add attributes to node that record the variable name and location in the TSG source which caused the creation of this node. This is useful for debugging.

The names for the attributes are provided by the user, so we don't clash with user attributes, or the user is able to use naming schemes that help further processing (something we do in tree-sitter-stack-graph).

Additionally, this changes the invocation of DSL execution to make it hopefully more resilient for future changes if such flags or options are introduced. Instead of passing separate parameters to `execute`, everything is grouped into an `ExecutionConfig` struct. Some values are required, and changing these will obviously break client code, but flags such as whether to use lazy execution are set by subsequent calls instead of required fields/parameters. As a result, there is not only a single public `execute` method, and lazy execution is called if the flag is set, instead of via a separate method.

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -60,12 +60,10 @@ impl File {
         &self,
         tree: &'tree Tree,
         source: &'tree str,
-        functions: &mut Functions,
-        globals: &Globals,
-        options: &ExecutionOptions,
+        config: &mut ExecutionConfig,
     ) -> Result<Graph<'tree>, ExecutionError> {
         let mut graph = Graph::new();
-        self.execute_into(&mut graph, tree, source, functions, globals, options)?;
+        self.execute_into(&mut graph, tree, source, config)?;
         Ok(graph)
     }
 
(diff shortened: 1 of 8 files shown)


src/execution.rs
```diff
@@ -129,6 +148,51 @@ impl File {
     }
 }
 
+/// Configuration for the execution of a File
+pub struct ExecutionConfig<'a, 'g> {
```
dcreager, line 152: All hail the builder pattern!

src/graph.rs
```diff
@@ -288,6 +288,10 @@ impl Attributes {
     {
         self.values.get(name.borrow())
     }
+
+    pub fn iter(&self) -> std::collections::hash_map::Iter<'_, Identifier, Value> {
```
dcreager, line 292: Same comment as in previous PR, you can use `impl Iterator<Item=&'_ Value> + '_` as the return type

dcreager (approved):

dcreager:
I really like the new `ExecutionConfig` pattern, that is much cleaner.

PR #50, opened 2022-03-17
Comprehensions

hendrikvanantwerpen:
This adds support for list comprehensions. At the moment it is very cumbersome to get the values of scoped variables for a list of nodes, or the source text of such a list. This allows you to write that down nicely without having to jump through hoops with mutable variables and functions:

```
(parameters (_)* @names)@params {
  let @params.names = [ (source-next name for name in @names ]
}
```

The comprehensions are subject to the same locality restrictions as for loops, so they don't make the language more powerful, just more convenient.

_Can be reviews commit-by-commit_

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -738,13 +738,10 @@ impl<'a> Parser<'a> {
             Ok(ast::ListLiteral { elements }.into())
         } else if let Ok(_) = self.consume_token(",") {
             self.consume_whitespace();
-            let other_elements = self.parse_sequence(']')?;
+            let mut elements = self.parse_sequence(']')?;
             self.consume_whitespace();
             self.consume_token("]")?;
-            let elements = vec![first_element]
-                .into_iter()
-                .chain(other_elements)
-                .collect();
+            elements.insert(0, first_element);
             Ok(ast::ListLiteral { elements }.into())
         } else {
             self.consume_token("for")?;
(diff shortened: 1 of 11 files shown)


src/parser.rs
```diff
@@ -725,19 +725,87 @@ impl<'a> Parser<'a> {
...
+            Ok(ast::ListLiteral { elements }.into())
+        } else if let Ok(_) = self.consume_token(",") {
+            self.consume_whitespace();
+            let other_elements = self.parse_sequence(']')?;
+            self.consume_whitespace();
+            self.consume_token("]")?;
+            let elements = vec![first_element]
+                .into_iter()
+                .chain(other_elements)
+                .collect();
```
dcreager, line 747: I think this is simpler as:

```suggestion
            let mut elements = self.parse_sequence(']')?;
            self.consume_whitespace();
            self.consume_token("]")?;
            elements.insert(0, first_element);
```

src/parser.rs
```diff
@@ -725,19 +725,87 @@ impl<'a> Parser<'a> {
...
+            Ok(ast::SetLiteral { elements }.into())
+        } else if let Ok(_) = self.consume_token(",") {
+            self.consume_whitespace();
+            let other_elements = self.parse_sequence('}')?;
+            self.consume_whitespace();
+            self.consume_token("}")?;
+            let elements = vec![first_element]
+                .into_iter()
+                .chain(other_elements)
+                .collect();
```
dcreager, line 789: ditto

src/reference/mod.rs
```diff
@@ -170,6 +172,25 @@
...
+//!
+//! Set comprehensions have similar syntax, but the resulting value will be a set instead of an ordered list:
+//!
+//! ``` tsg
+//! { (some-function x) for x in @xs }
+//! { @x.scoped_var for x in @xs }
+//! ```
+//!
+//! List and set comprehensions are subject to the same restrictions as for loops, which means the list
+//! value that is iterated over must be local.  It is therefore not possible to iterator over the value
```
dcreager, line 191:
```suggestion
//! value that is iterated over must be local.  It is therefore not possible to iterate over the value
```

dcreager (approved):
Very nice

PR #52, opened 2022-03-21
Add `TreeWithParseError*` types to safely move tree and parse errors

hendrikvanantwerpen:
**The implementation uses `unsafe` code, so careful review is appreciated.**

1. This adds types that allow safely moving a `Tree` and `ParseError`s (which contain `Node<'tree>` references) together. This is normally not possible in Rust because moving the `Tree` would invalidate the references, but in the case of tree-sitter, the references point not to the Rust value, but to C allocated memory. This memory is not moved when the `Tree` is, and thus the references remain valid.

   The `TreeWithParseError*` types use `unsafe` code to erase the lifetime `'tree` to `static`. The `Tree` is owned by the value, and is only dropped when the value is. When the `Tree` is dropped, the memory is deallocated, so the parse errors should not outlive the `Tree`. This is ensured because the parse errors are only available through a reference, which cannot outlive the value they are tied to (which is the `TreeWithParseError*` that ownes the corresponding `Tree`).

2. The parse error formatting has been changed to _not_ include the position of the error. This gives the user more flexibility in how that should be formatted. E.g., in a UI, the position may not even be included in the message, or a tool may use a particular formatting convention for source positions.

_Can be reviewed commit-by-commit._

### DIFF
--- a/src/parse_error.rs
+++ b/src/parse_error.rs
@@ -52,6 +52,13 @@ impl<'tree> ParseError<'tree> {
 }
 
 impl<'tree> ParseError<'tree> {
+    pub fn node(&self) -> &Node<'tree> {
+        match self {
+            Self::Missing(node) => node,
+            Self::Unexpected(node) => node,
+        }
+    }
+
     pub fn display(&self, source: &'tree str, verbose: bool) -> ParseErrorDisplay {
         ParseErrorDisplay {
             error: self,
@@ -81,7 +88,6 @@ impl std::fmt::Display for ParseErrorDisplay<'_> {
         };
         let line = node.start_position().row;
         let start_column = node.start_position().column;
-        write!(f, " on line {} column {}", line + 1, start_column + 1)?;
         if node.byte_range().is_empty() {
             writeln!(f, "")?;
         } else {
--- a/src/parse_error.rs
+++ b/src/parse_error.rs
@@ -31,9 +31,7 @@ impl<'tree> ParseError<'tree> {
     /// This returns a type, combining the tree and the error, that can be moved safely,
     /// which is not possible with a seperate tree and error.
     pub fn into_first(tree: Tree) -> TreeWithParseErrorOption {
-        let mut errors = Vec::new();
-        find_errors(&tree, &mut errors, true);
-        crate::new_tree_with_parse_error_option!(tree, errors.into_iter().next())
+        TreeWithParseErrorOption::into_first(tree)
     }
 
     /// Return all parse errors in the given tree.
@@ -47,9 +45,7 @@ impl<'tree> ParseError<'tree> {
     /// This returns a type, combining the tree and the errors, that can be moved safely,
     /// which is not possible with a seperate tree and errors.
     pub fn into_all(tree: Tree) -> TreeWithParseErrorVec {
-        let mut errors = Vec::new();
-        find_errors(&tree, &mut errors, false);
-        crate::new_tree_with_parse_error_vec!(tree, errors)
+        TreeWithParseErrorVec::into_all(tree)
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/parse_error.rs
```diff
@@ -138,3 +124,195 @@ impl std::fmt::Display for ParseErrorDisplay<'_> {
...
+
+#[macro_export]
+macro_rules! new_tree_with_parse_error_option {
+    ($tree:expr,$error:expr) => {{
+        TreeWithParseErrorOption {
+            error: unsafe { std::mem::transmute($error) },
+            tree: $tree,
+        }
+    }};
+}
```
dcreager, line 241: I follow your explanation of why it's safe to bundle the `Tree` and its `Node`s and agree with it.  This part gives me pause, though — I think these macros open up too much of a back door on the unsafe code.  Right now anyone* can call the macro and it's not obvious at the call site that unsafe behavior is happening.  Ideally we could write this as a function, so that we could at least verify that `$error` is a valid instead of `ParseError` — but we can't do that, since you'd have to pass in the `Tree` and something containing a reference to that `Tree` as parameters.

_*Not really "anyone", since the macro is filling in some package-local fields, but I think the argument still applies._

Could you instead move the `into_first` method to belong to `TreeWithParseErrorOption`?  Use exactly the same logic to walk the tree and build up the vector, but use this same transmute logic to drop the lifetime.  But you could do it directly, instead of via a call to this macro.

(You could still have the _public_ `into_first` method to live where it currently is, if you want — the proposed one here would then be package-local.)

hendrikvanantwerpen: I see how users cannot really use this macro, since it expands at use site and is checked in the context of the expansion 🤦🏻. I think your suggestion is good, to move it into a function of the type itself.

I was looking for a way to give users the possibility to do the same, but I now think that might not be possible in a safe way. Making the fields public certainly is unsound, because it exposes the `Node` with a `static` lifetime.

dcreager: > I was looking for a way to give users the possibility to do the same, but I now think that might not be possible in a safe way.

Yeah I think that's a change that better belongs up in tree-sitter itself — to make `Tree` and `Node` have lifetimes that don't require these kinds of shenanigans in the first place.

> Making the fields public certainly is unsound, because it exposes the `Node` with a `static` lifetime.

Yes good point, I hadn't considered that!  You're right that that also wouldn't work

src/parse_error.rs
```diff
@@ -138,3 +120,197 @@ impl std::fmt::Display for ParseErrorDisplay<'_> {
...
+impl std::fmt::Debug for TreeWithParseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{:?}", self.error)
+    }
+}
+
+// Send and Sync must be implemented for ParseError -> Node -> ffi::TSTree
+// This is okay because Send and Sync _are_ implemented for Tree, which also holds ffi::TSTree
+unsafe impl Send for TreeWithParseError {}
+unsafe impl Sync for TreeWithParseError {}
```
dcreager, line 220: It doesn't look like this type is used anywhere, can we drop it?  (If you'd rather keep it, remove the macro like you did with the types below)

dcreager (approved):

src/parse_error.rs
hendrikvanantwerpen: Forgot about that one, sorry about that. It is used in `TreeWithParseErrorOption::into_option`, which returns an `Option<TreeWithParseError`. It is a bit cumbersome to have three such similar types, but since users are stuck with what we provide them, I added this for completeness. I'll remove the macro.

src/parse_error.rs
hendrikvanantwerpen: > Yeah I think that's a change that better belongs up in tree-sitter itself - to make Tree and Node have lifetimes that don't require these kinds of shenanigans in the first place.

Agreed. I was wondering though if that is possible. The `Tree` deallocates memory when it is dropped. If you have types `Tree<'a>` and `Node<'a>`, then (a) where would `'a` for the `Tree` come from, and (b) could `'a` be a longer then the lifetime of `Tree<'a>`?

PR #54, opened 2022-04-12
Serialize integers/node/graph IDs & lists/sets distinctly

### LINKED ISSUE #46: Can't distinguish integers, graph node IDs, and syntax node IDs in JSON output format
Right now we just spit out numbers for all three.

robrix:
- [x] Fixes #46.
- [x] Does the same for lists/sets.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -529,7 +529,12 @@ impl Serialize for Value {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         match self {
             Value::Null => serializer.serialize_none(),
-            Value::Boolean(value) => serializer.serialize_bool(*value),
+            Value::Boolean(bool) => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "bool")?;
+                map.serialize_entry("bool", bool)?;
+                map.end()
+            }
             Value::Integer(int) => {
                 let mut map = serializer.serialize_map(None)?;
                 map.serialize_entry("type", "int")?;
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -530,7 +530,12 @@ impl Serialize for Value {
         match self {
             Value::Null => serializer.serialize_none(),
             Value::Boolean(value) => serializer.serialize_bool(*value),
-            Value::Integer(value) => serializer.serialize_u32(*value),
+            Value::Integer(int) => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "int")?;
+                map.serialize_entry("int", int)?;
+                map.end()
+            }
             Value::String(str) => {
                 let mut map = serializer.serialize_map(None)?;
                 map.serialize_entry("type", "string")?;
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -547,13 +547,13 @@ impl Serialize for Value {
             Value::SyntaxNode(node) => {
                 let mut map = serializer.serialize_map(None)?;
                 map.serialize_entry("type", "syntaxNode")?;
-                map.serialize_entry("value", &node.index)?;
+                map.serialize_entry("id", &node.index)?;
                 map.end()
             }
             Value::GraphNode(node) => {
                 let mut map = serializer.serialize_map(None)?;
                 map.serialize_entry("type", "graphNode")?;
-                map.serialize_entry("value", &node.0)?;
+                map.serialize_entry("id", &node.0)?;
                 map.end()
             }
         }
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -532,7 +532,6 @@ impl Serialize for Value {
             Value::Boolean(value) => serializer.serialize_bool(*value),
             Value::Integer(value) => serializer.serialize_u32(*value),
             Value::String(value) => serializer.serialize_str(value),
-            // FIXME: there's no way to distinguish sets and lists, so we can't roundtrip accurately
             Value::List(list) => {
                 let mut map = serializer.serialize_map(None)?;
                 map.serialize_entry("type", "list")?;
@@ -545,7 +544,6 @@ impl Serialize for Value {
                 map.serialize_entry("value", set)?;
                 map.end()
             }
-            // FIXME: we don't distinguish between syntax tree node IDs, graph node IDs, and integers
             Value::SyntaxNode(node) => {
                 let mut map = serializer.serialize_map(None)?;
                 map.serialize_entry("type", "syntaxNode")?;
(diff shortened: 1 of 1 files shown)


src/graph.rs
```diff
@@ -532,12 +532,30 @@ impl Serialize for Value {
             Value::Boolean(value) => serializer.serialize_bool(*value),
             Value::Integer(value) => serializer.serialize_u32(*value),
             Value::String(value) => serializer.serialize_str(value),
```
robrix, line 534: Maybe we should serialize these as hashes too?

src/graph.rs
```diff
@@ -532,12 +532,30 @@ impl Serialize for Value {
...
-            Value::Set(set) => serializer.collect_seq(set),
-            // FIXME: we don't distinguish between syntax tree node IDs, graph node IDs, and integers
-            Value::SyntaxNode(node) => serializer.serialize_u32(node.index),
-            Value::GraphNode(node) => serializer.serialize_u32(node.0),
+            Value::List(list) => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "list")?;
+                map.serialize_entry("value", list)?;
+                map.end()
+            }
```
robrix, line 540: The format seems unnecessarily sparse, but it matches what we've done elsewhere. If we were to tighten it up, I'd suggest `{ "list": […] }`.

robrix: It also seems like we could leave this one using arrays and only give sets special formatting.

src/graph.rs
```diff
@@ -532,12 +532,30 @@ impl Serialize for Value {
...
+            Value::SyntaxNode(node) => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "syntaxNode")?;
+                map.serialize_entry("value", &node.index)?;
+                map.end()
+            }
+            Value::GraphNode(node) => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "graphNode")?;
+                map.serialize_entry("value", &node.0)?;
```
hendrikvanantwerpen, line 556: In a similar vain, I'd suggest `id` here if the graph nodes have an `id` field, or `index` if there live in an array and their position is used to identify them.

hendrikvanantwerpen (commented):
Thanks for addressing this! I left comments on your own questions and a few questions of my own.

Since JSON was already included in a released version, you may also want to update the CHANGELOG for these changes.

src/graph.rs
```diff
@@ -532,12 +532,30 @@ impl Serialize for Value {
...
+            Value::List(list) => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "list")?;
+                map.serialize_entry("value", list)?;
+                map.end()
+            }
+            Value::Set(set) => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "set")?;
+                map.serialize_entry("value", set)?;
```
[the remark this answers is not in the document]
robrix: good call 👍🏻

src/graph.rs
robrix: Uniformity, basically: serializing all of the different types of `Value` the same way. I rephrase the question as: how do we value uniformity of `Value` serializations relative to sticking close to JSON's primitives where possible?

src/graph.rs
robrix: I don't _think_ further fields are likely, as those would likely be annotated via attrs on nodes/edges rather than on the values making them up. (I think?) But it's a point to consider.

robrix:
There are two major questions remaining:

1. is it better to distinguish different kinds of values via a `type` field, like this PR currently does (sparse), or to use a denser encoding where the presence of a specific field determines its type? e.g.:

    ```json
    { "type": "graphNode"
    , "value": 1 }
    ```

    vs.

    ```json
    { "graphNode": 1 }
    ```

    Apart from that there's uniformity of the key for the actual value, but I agree with @hendrikvanantwerpen that the key for lists/sets is better as `values` than `value`, if we go with sparse encodings or whatever.

3. should we uniformly encode _all_ values as hashes/objects/dictionaries/whatever they're called, or should we encode the ones most obviously related to JSON primitives… primitively? e.g.

    ```json
    "arbitrary text"
    ```

    vs.

    ```json
    { "type": "string"
    , "value": "arbitrary text" }
    ```

cc @tree-sitter/semantic-code

hendrikvanantwerpen:
Re 1. My impression is that the sparse representation is more idiomatic of JSON than the dense one. As @patrickt pointed out, the sparse representation also makes it easier to query with `jq`.

src/graph.rs
```diff
@@ -528,16 +528,53 @@ impl std::fmt::Display for Value {
...
-            // FIXME: there's no way to distinguish sets and lists, so we can't roundtrip accurately
-            Value::List(list) => serializer.collect_seq(list),
-            Value::Set(set) => serializer.collect_seq(set),
-            // FIXME: we don't distinguish between syntax tree node IDs, graph node IDs, and integers
-            Value::SyntaxNode(node) => serializer.serialize_u32(node.index),
-            Value::GraphNode(node) => serializer.serialize_u32(node.0),
+            Value::Null => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "null")?;
+                map.end()
```
robrix, line 534: The sparse representation for `Value::Null` is unique in not having any kind of value field.

src/graph.rs
```diff
@@ -528,16 +528,53 @@ impl std::fmt::Display for Value {
...
+            }
+            Value::Boolean(bool) => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "bool")?;
+                map.serialize_entry("bool", bool)?;
+                map.end()
+            }
```
robrix, line 541: `Value::Boolean`, `Value::Integer`, and `Value::String` use their `type` field's value as the value field's key. I don't think we can get a lot more semantic with this, unfortunately.

src/graph.rs
```diff
@@ -528,16 +528,53 @@ impl std::fmt::Display for Value {
...
+            Value::SyntaxNode(node) => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "syntaxNode")?;
+                map.serialize_entry("id", &node.index)?;
+                map.end()
+            }
+            Value::GraphNode(node) => {
+                let mut map = serializer.serialize_map(None)?;
+                map.serialize_entry("type", "graphNode")?;
+                map.serialize_entry("id", &node.0)?;
```
robrix, line 575: Per @hendrikvanantwerpen's suggestion, I've named this `id` to line up with graph nodes' `id` field.

robrix (commented):
Ready for (re-)review.

hendrikvanantwerpen (approved):
I think this is fine! This worked out a bit different than I expected, and I am doubting how good my original suggestions actually were. But I feel like I've moved this discussion way past the ship-to-learn point, so I'm approving this.

PR #59, opened 2022-05-10
Add VS Code extension for tree-sitter-graph syntax

hendrikvanantwerpen:
This adds a VS Code extension to provide syntax highlighting for tree-sitter-graph files.

Here's an impression (the exact colors are not determined by this extension, but, I suppose, by my VS Code theme):

image

I would also like to publish this in the VS Code Marketplace.

### DIFF
--- a/vscode/syntaxes/tree-sitter-graph.tmLanguage.json
+++ b/vscode/syntaxes/tree-sitter-graph.tmLanguage.json
@@ -1,0 +1,52 @@
+{
+	"$schema": "https://raw.githubusercontent.com/martinring/tmlanguage/master/tmlanguage.json",
+	"name": "Tree-Sitter Graph",
+	"patterns": [{
+		"include": "#keywords"
+	}, {
+		"include": "#functions"
+	}, {
+		"include": "#constants"
+	}, {
+		"include": "#strings"
+	}, {
+		"include": "#comments"
+	}],
+	"repository": {
+		"keywords": {
+			"patterns": [{
+				"name": "keyword.control.tsg",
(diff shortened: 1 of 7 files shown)


vscode/syntaxes/tree-sitter-graph.tmLanguage.json
```diff
@@ -0,0 +1,52 @@
...
+		"keywords": {
+			"patterns": [{
+				"name": "keyword.control.tsg",
+				"match": "^\\s*(attr|attribute|edge|for|global|if|let|node|none|print|scan|set|some|var)\\b"
+			}]
+		},
+		"functions": {
+			"patterns": [{
+				"name": "support.function.tsg",
+				"match":"(?<=\\(\\s*)(and|end-column|end-row|is-empty|is-null|length|named-child-count|named-child-index|node|node-type|not|or|plus|replace|source-text|start-column|start-row)\\b"
```
rewinfrey, line 25: It's been awhile since I used a positive lookbehind in a regex.  Just to double check my understanding,  this will match only if there's a preceding `(` before one of the support function names (e.g. `(and` would match but not `and`)?

rewinfrey (approved):
This is really exciting!  ✨ I'm not sure I can comment on the VS Code marketplace,  but I'm not aware of any reasons why it cannot be added.

vscode/syntaxes/tree-sitter-graph.tmLanguage.json
hendrikvanantwerpen: Exactly! It is not prefect, but at least identifiers with the same name as functions are not highlighted incorrectly. It would still incorrectly highlight tree nodes with the same name as those functions. However, fixing that with textmate is (I think) painful, so for now I don't want to spend the effort for that.

PR #78, opened 2022-07-28
Pass globals as command-line options

### LINKED ISSUE #76: Supply globals from the CLI
I'm working on a `.tsg` program which is eventually going to be run from some specialized process or other which may or may not already exist. In the meantime, however, I'd like to be able to iterate on it and test it against real sources without having to do that development effort, and enjoy the benefits of `tree-sitter-graph`'s other features (JSON output, error presentation, etc.).

I can do this already, but only up to a point: globals which the eventual host process may provide are of course not available on the CLI—but they could be!

I propose adding a `--global NAME=VALUE` option which can be supplied zero or more times, taking a global variable name and the value to give it. We'd of course need to do something about types to distinguish (for example) ints from strings, which might become intricate given composites like lists (and lists of lists, and…).

robrix:
- [x] Fixes #76.
- ~~How should types other than `String` be handled?~~ See #86.
- ~~`String::split_once()` seems not to match when the rhs (after the delimiter) is empty. How should we handle empty globals, e.g. empty strings?~~ See #87.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ cli = ["clap", "env_logger", "tree-sitter-config", "tree-sitter-loader"]
 
 [dependencies.clap]
 optional = true
-version = "2.32"
+version = "3.2"
 
 [dependencies.env_logger]
 optional = true
--- a/src/bin/tree-sitter-graph/main.rs
+++ b/src/bin/tree-sitter-graph/main.rs
@@ -36,13 +36,13 @@ fn main() -> Result<()> {
         .arg(Arg::with_name("source").index(2).required(true))
         .arg(
             Arg::with_name("quiet")
-                .short("q")
+                .short('q')
                 .long("quiet")
                 .help("Suppress console output"),
         )
         .arg(
             Arg::with_name("lazy")
-                .short("z")
+                .short('z')
                 .long("lazy")
                 .help("Use lazy evaluation (experimental)"),
         )
@@ -50,7 +50,7 @@ fn main() -> Result<()> {
         .arg(Arg::with_name("json").long("json").takes_value(false))
         .arg(
             Arg::with_name("output")
-                .short("o")
+                .short('o')
                 .long("output")
                 .requires("json")
                 .takes_value(true),
--- a/src/bin/tree-sitter-graph/main.rs
+++ b/src/bin/tree-sitter-graph/main.rs
@@ -10,6 +10,7 @@ use std::path::Path;
 use anyhow::anyhow;
 use anyhow::Context as _;
 use anyhow::Result;
+use clap::builder::ArgAction;
 use clap::App;
 use clap::Arg;
 use tree_sitter::Parser;
@@ -60,6 +61,12 @@ fn main() -> Result<()> {
                 .long("allow-parse-errors")
                 .takes_value(false),
         )
+        .arg(
+            Arg::with_name("global")
+                .long("global")
+                .takes_value(true)
+                .action(ArgAction::Append),
+        )
         .get_matches();
 
     let tsg_path = Path::new(matches.value_of("tsg").unwrap());
@@ -67,6 +74,7 @@ fn main() -> Result<()> {
     let current_dir = std::env::current_dir().unwrap();
     let quiet = matches.is_present("quiet");
     let lazy = matches.is_present("lazy");
+    let globals = matches.get_many::<String>("global").unwrap_or_default();
 
     let config = Config::load()?;
     let mut loader = Loader::new()?;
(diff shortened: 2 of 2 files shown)


robrix:
cc @BekaValentine — this might be convenient for you.

robrix:
I'm ok with leaving both the "how do we handle non-`String`?" and "how do we supply empty strings?" questions as future work. This PR is already useful as-is and we can extend later. Furthermore, I think the current behaviour doesn't prevent us from resolving either of those questions. If folks are ok with this, let's see about getting it merged in.

Cargo.toml
```diff
@@ -46,7 +46,7 @@ cli = ["clap", "env_logger", "tree-sitter-config", "tree-sitter-loader"]
 
 [dependencies.clap]
 optional = true
-version = "2.32"
+version = "3.2"
```
robrix, line 49: Had to bump this to be able to supply the args the way I wanted.

src/bin/tree-sitter-graph/main.rs
```diff
@@ -36,21 +39,21 @@ fn main() -> Result<()> {
         .arg(Arg::with_name("source").index(2).required(true))
         .arg(
             Arg::with_name("quiet")
-                .short("q")
+                .short('q')
```
robrix, line 42: Consequence of bumping `clap`.

src/bin/tree-sitter-graph/main.rs
```diff
@@ -60,13 +63,30 @@ fn main() -> Result<()> {
...
+    let mut globals_ = Variables::new();
+    for kv in globals {
+        let kv_ = kv
+            .split_once('=')
+            .with_context(|| format!("Expected key-value pair separated by '=', got {}.", kv))?;
+        globals_.add(
+            Identifier::from(kv_.0),
+            graph::Value::String(kv_.1.to_string()),
+        )?;
+    }
```
robrix, line 89: I would have preferred to do this with a single map but I couldn't figure out how to express the two nested levels of failure (failure in parsing global values, and failure in the surrounding context). If anyone has suggestions, I'm all ears; I think we should be able to avoid both the second variable and the mutability, and (ideally) end up with something more readable as well.

src/bin/tree-sitter-graph/main.rs
```diff
@@ -116,8 +136,7 @@ fn main() -> Result<()> {
     }
 
     let mut functions = Functions::stdlib();
-    let globals = Variables::new();
-    let mut config = ExecutionConfig::new(&mut functions, &globals).lazy(lazy);
+    let mut config = ExecutionConfig::new(&mut functions, &globals_).lazy(lazy);
```
robrix, line 139: This is the key to how this all works: instead of passing in an empty map of `Variables`, we pass in the one we built above instead. Really quite tidy, that.

robrix (commented):
Ready for review; see above for comment about non-string-valued globals & empty strings. If we're ok with leaving them as future work, I'll file issues when merging this PR.

src/bin/tree-sitter-graph/main.rs
dcreager: I actually like how you've done it here!  As you point out, it becomes messy with the two levels of error handling, and I think this rendering makes the control flow clearer.

(It's doable, and probably involves the [`collect` method](https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect), which takes a vec of results and turns it into a result of a vec.  But again, I'd rather we keep it as you've written it here.)

dcreager (approved):
This looks great to me!

src/bin/tree-sitter-graph/main.rs
dcreager: (Also, if you want to bound the region where `globals_` is mutable, you can do something like this:

``` rust
let globals_ = globals_;
```

after you're done mutating it.  That makes it immutable for the remainder of the scope.)

robrix: > I actually like how you've done it here! As you point out, it becomes messy with the two levels of error handling, and I think this rendering makes the control flow clearer.
> 
> (It's doable, and probably involves the [`collect` method](https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect), which takes a vec of results and turns it into a result of a vec. But again, I'd rather we keep it as you've written it here.)

Cool, if this is idiomatic then I'll be happy to :shipit:

> (Also, if you want to bound the region where `globals_` is mutable, you can do something like this:
> 
> ```rust
> let globals_ = globals_;
> ```
> 
> after you're done mutating it. That makes it immutable for the remainder of the scope.)

Oh, that's a lovely use of shadowing! Thank you.

robrix:
Filed #86 & #87.

hendrikvanantwerpen:
Nice improvement @robrix!

PR #91, opened 2022-08-11
Print excerpts from tsg & source files on error

### LINKED ISSUE #63: Include source context in errors
`tree-sitter-graph` errors (for when you don't have a rule to cover some piece of syntax) conveniently show you the node type and range in errors:

```
Error: Cannot execute TSG file ./semantic-analysis/python.tsg

Caused by:
    0: Executing for child in @children { ... } at (68, 3)
    1: Undefined variable child.node on node [syntax node expression_statement (3, 5)]
```

It would be neat to extend that with a bit of the source, at least the first line or so (and maybe the preceding and following one or two for context?).

### LINKED ISSUE #83: Show source for rule file locations
When we show an error indicating a given location in a rule file, I'd like to see an excerpt of the source around that location. One such should be shown for each rule file location involved in the error (tho maybe we should filter for duplicates).

I don't know if we should also show an excerpt for non-error locations, e.g. "Executing edge … -> … at (x, y)" entries. Perhaps if they're the last one before a source file (syntax node) location?

### LINKED ISSUE #84: Show source excerpts for source file locations
Errors involving syntax nodes reference locations in the source file. We should show excerpts from the source file, just like #83 discusses for rules files.

See also #82 re: ensuring we include those locations in more errors, and #81 re: tracking multiple locations for duplicates specifically.

### LINKED ISSUE #90: Show where the rule matched for dynamic errors
I've got a duplicate edge error, which is discovered dynamically when running against some source file. How can this edge already exist? What node is it actually talking about? Who knows!

I'd like the error to include:

1. Where the match in the source file began, along with an excerpt of (at least) the first line of that source (#84).
2. Where the bound nodes were within the source, e.g. by picking a terminal-usable colour for each variable, drawing the variable's name in that, and drawing the corresponding span of the excerpt in that as well, with child nodes taking precedence over parent nodes colour-wise.
3. What variables exist on the various nodes. (Maybe just the ones we've explicitly bound, although a REPL (#77) might allow one to get more.)

robrix:
- [x] Fixes #63.
- [x] Fixes #83.
- [x] Fixes #84.
- [x] Fixes #90.
- [x] Apply to lazy evaluation.

screenshot of terminal session showing new error messages with context from both tsg and target language files

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ thiserror = "1.0"
 tree-sitter = "0.20"
 serde = "1.0"
 serde_json = "1.0"
+ansi_term = "0.12"
 
 [dependencies.string-interner]
 version = "0.12"
--- a/src/bin/tree-sitter-graph/main.rs
+++ b/src/bin/tree-sitter-graph/main.rs
@@ -139,7 +139,7 @@ fn main() -> Result<()> {
     let functions = Functions::stdlib();
     let mut config = ExecutionConfig::new(&functions, &globals_).lazy(lazy);
     let graph = file
-        .execute(&tree, &source, &mut config, &NoCancellation)
+        .execute(&tree, &source, &tsg, &mut config, &NoCancellation)
         .with_context(|| format!("Cannot execute TSG file {}", tsg_path.display()))?;
 
     let json = matches.is_present("json");
--- a/src/bin/tree-sitter-graph/main.rs
+++ b/src/bin/tree-sitter-graph/main.rs
@@ -139,7 +139,15 @@ fn main() -> Result<()> {
     let functions = Functions::stdlib();
     let mut config = ExecutionConfig::new(&functions, &globals_).lazy(lazy);
     let graph = file
-        .execute(&tree, &source, &tsg, &mut config, &NoCancellation)
+        .execute(
+            &tree,
+            source_path,
+            &source,
+            tsg_path,
+            &tsg,
+            &mut config,
+            &NoCancellation,
+        )
         .with_context(|| format!("Cannot execute TSG file {}", tsg_path.display()))?;
 
     let json = matches.is_present("json");
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -54,6 +54,7 @@ use crate::variables::Globals;
 use crate::variables::VariableMap;
 use crate::variables::Variables;
 use crate::Identifier;
+use crate::Location;
 
 impl File {
     /// Executes this graph DSL file against a source file.  You must provide the parsed syntax
@@ -252,8 +253,8 @@ pub enum ExecutionError {
     EmptyRegexCapture(String),
     #[error("Undefined edge {0}")]
     UndefinedEdge(String),
-    #[error("Undefined variable {0}")]
-    UndefinedVariable(String),
+    #[error("Undefined variable {0} at {1}")]
+    UndefinedVariable(String, Location),
     #[error("Cannot add scoped variable after being forced {0}")]
     VariableScopesAlreadyForced(String),
     #[error(transparent)]
@@ -821,10 +822,10 @@ impl ScopedVariable {
         if let Some(value) = variables.get(&self.name) {
             Ok(value)
         } else {
-            Err(ExecutionError::UndefinedVariable(format!(
-                "{} on node {}",
-                self, scope,
-            )))
+            Err(ExecutionError::UndefinedVariable(
+                format!("{} on node {}", self, scope,),
+                scope.location(),
+            ))
         }
     }
 
@@ -875,7 +876,7 @@ impl UnscopedVariable {
         } else {
             exec.locals.get(&self.name)
         }
-        .ok_or_else(|| ExecutionError::UndefinedVariable(format!("{}", self)))
+        .ok_or_else(|| ExecutionError::UndefinedVariable(format!("{}", self), self.location))
     }
 
     fn add(
@@ -906,7 +907,7 @@ impl UnscopedVariable {
             if exec.locals.get(&self.name).is_some() {
                 ExecutionError::CannotAssignImmutableVariable(format!("{}", self))
             } else {
-                ExecutionError::UndefinedVariable(format!("{}", self))
+                ExecutionError::UndefinedVariable(format!("{}", self), self.location)
             }
         })
     }
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -393,7 +393,7 @@ impl Stanza {
                         format!(
                             "Executing {} {}",
                             statement,
-                            Excerpt::from_source(source, statement.location())
+                            Excerpt::from_source(tsg_source, statement.location())
                         )
                     })
                 })?;
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -386,7 +386,17 @@ impl<'a> std::fmt::Display for Excerpt<'a> {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         write!(
             f,
-            "{}{}{}\n{}{}{}{}",
+            "{}:{}:{}:\n{}{}{}\n{}{}{}{}",
+            // path and line/col
+            Colour::White
+                .bold()
+                .paint(self.path.to_str().unwrap_or("<unknown file>")),
+            Colour::White
+                .bold()
+                .paint(format!("{}", self.location.row + 1)),
+            Colour::White
+                .bold()
+                .paint(format!("{}", self.location.column + 1)),
             // first line: line number & source
             Colour::Blue.paint(format!("{}", self.location.row + 1)),
             Colour::Blue.paint(" | "),
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -929,9 +929,10 @@ impl ScopedVariable {
             .add(self.name.clone(), value, mutable)
             .map_err(|_| {
                 ExecutionError::DuplicateVariable(format!(
-                    "{}\n{}",
+                    "{}\n{}\n{}",
                     self,
-                    Excerpt::from_source(exec.tsg_source, self.location)
+                    Excerpt::from_source(exec.tsg_source, self.location),
+                    Excerpt::from_source(exec.source, scope.location())
                 ))
             })
     }
@@ -950,9 +951,10 @@ impl ScopedVariable {
         let variables = exec.scoped.get(scope);
         variables.set(self.name.clone(), value).map_err(|_| {
             ExecutionError::DuplicateVariable(format!(
-                "{}\n{}",
+                "{}\n{}\n{}",
                 self,
-                Excerpt::from_source(exec.tsg_source, self.location)
+                Excerpt::from_source(exec.tsg_source, self.location),
+                Excerpt::from_source(exec.source, scope.location())
             ))
         })
     }
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -353,15 +353,14 @@ impl<'a> std::fmt::Display for Excerpt<'a> {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         let gutter_width = self.gutter_width();
         let column_padding = self.location.column;
+        let empty = "";
         write!(
             f,
-            "{}{}{}\n{:gutter_width$}{}{:column_padding$}{}",
+            "{}{}{}\n{empty:gutter_width$}{}{empty:column_padding$}{}",
             Colour::Blue.paint(format!("{}", self.location.row + 1)),
             Colour::Blue.paint(" | "),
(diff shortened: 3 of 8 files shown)


Cargo.toml
```diff
@@ -24,6 +24,7 @@ thiserror = "1.0"
 tree-sitter = "0.20"
 serde = "1.0"
 serde_json = "1.0"
+ansi_term = "0.12"
```
robrix, line 27: I was going to say something glib about terminal colourization being _so important_, but it actually kind of is, as an accessibility thing. We highlight very little, but breaking it up into different colours makes it much easier to immediately bucket them appropriately in your head and focus your attention where you need it next.

src/bin/tree-sitter-graph/main.rs
```diff
@@ -138,7 +138,7 @@ fn main() -> Result<()> {
     let mut functions = Functions::stdlib();
     let mut config = ExecutionConfig::new(&mut functions, &globals_).lazy(lazy);
     let graph = file
-        .execute(&tree, &source, &mut config)
+        .execute(&tree, source_path, &source, tsg_path, &tsg, &mut config)
```
robrix, line 141: Showing excerpts requires the path to the file, the contents of the file, and a location. We get the location dynamically, and the other two are passed as static parts of the execution context.

Finally, we show excerpts from both the .tsg file and the target file, so we need to pass both in.

src/execution.rs
```diff
@@ -112,7 +148,10 @@ impl File {
         for stanza in &self.stanzas {
             stanza.execute(
                 tree,
+                source_path,
                 source,
+                tsg_path,
+                tsg_source,
```
robrix, line 154: I had a quick go at bundling up both path and source into a `struct`, but ultimately didn't feel it was worth the space.

src/execution.rs
```diff
@@ -320,16 +411,51 @@ impl Stanza {
...
+                            Excerpt::from_source(tsg_path, tsg_source, self.location),
+                            node.kind(),
+                            Excerpt::from_source(
+                                source_path,
+                                source,
+                                Location::from(node.range().start_point)
+                            )
+                        )
+                    })
+                })?;
```
robrix, line 435: I had initially toyed with adding a `Location` to each kind of error—or actually two, one for .tsg, one for target. Then I had each individual error site add an `Excerpt` to its description with whatever .tsg/source locations we could find.

I ended up adding them once and for all here, because it's _much_ simpler than adding locations and such to every kind of error, and unlike displaying excerpts at each error site, it's automatically exhaustive. However, it does come with the caveat that the precision of the error location is reduced relative to the error site information; e.g. for a duplicate scoped variable `node @expr.thing`, we get a caret pointing at `thing`; with the above approach we instead point at the start of the line. I think this was a worthwhile tradeoff, tho still potentially confusing in some cases.

src/execution.rs
```diff
@@ -320,16 +411,51 @@ impl Stanza {
...
+                        let node = mat
+                            .captures
+                            .iter()
+                            .find(|c| c.index as usize == self.full_match_file_capture_index)
+                            .unwrap()
+                            .node;
+                        format!(
+                            "While executing statement {}\n{}\nin stanza\n{}\nmatching ({}) node\n{}",
+                            statement,
+                            Excerpt::from_source(tsg_path, tsg_source, statement.location()),
```
robrix, line 425: We're using `Location`s instead of `Range`s (that is, these were already in use prior to this PR), so we only print the one line. Future work if we want `clang`-style diagnostics, for example.

src/execution.rs
```diff
@@ -320,16 +411,51 @@ impl Stanza {
...
+                            Excerpt::from_source(tsg_path, tsg_source, self.location),
```
robrix, line 426: Even if `Excerpt`s did support `Range`s, we don't plumb the range of the query around, so just printing its first line is a better option than e.g. printing the entire stanza.

src/execution.rs
```diff
@@ -320,16 +411,51 @@ impl Stanza {
...
             for statement in &self.statements {
-                statement
-                    .execute(&mut exec)
-                    .with_context(|| format!("Executing {}", statement))?;
+                statement.execute(&mut exec).or_else(|e| {
+                    Err(e).with_context(|| {
+                        let node = mat
+                            .captures
+                            .iter()
+                            .find(|c| c.index as usize == self.full_match_file_capture_index)
```
robrix, line 419: 🎩 @hendrikvanantwerpen for a) telling me about this, and b) making this possible in the first place.

src/execution.rs
```diff
@@ -291,11 +330,63 @@ impl<'a> ScopedVariables<'a> {
...
+
 impl Stanza {
-    fn execute<'g, 'l, 's, 'tree>(
+    fn execute<'a, 'g, 'l, 's, 'tree>(
         &self,
         tree: &'tree Tree,
+        source_path: &'tree Path,
         source: &'tree str,
+        tsg_path: &'a Path,
+        tsg_source: &'a str,
```
robrix, line 389: In a past life (one commit ago at time of writing), these were all added to the `ExecutionContext` along with `source`. Happily, since we're contextualizing the errors in this very function, we don't need to pass them around to every `evaluate` definition.

(`source` still has to be carried around in order to evaluate calls.)

src/execution.rs
```diff
@@ -320,16 +411,51 @@ impl Stanza {
...
+            Statement::CreateGraphNode(s) => s.location,
+            Statement::AddGraphNodeAttribute(s) => s.location,
+            Statement::CreateEdge(s) => s.location,
+            Statement::AddEdgeAttribute(s) => s.location,
+            Statement::Scan(s) => s.location,
+            Statement::Print(s) => s.location,
+            Statement::If(s) => s.location,
+            Statement::ForIn(s) => s.location,
+        }
+    }
```
robrix, line 457: I feel like there ought to be a better way to do this sort of thing, but if there is, I don't know it.

robrix (commented):
Ready for review.

src/execution.rs
```diff
@@ -54,19 +56,31 @@ use crate::variables::Globals;
...
     /// tree (`tree`) as well as the source text that it was parsed from (`source`).  You also
     /// provide the set of functions and global variables that are available during execution.
-    pub fn execute<'tree>(
+    pub fn execute<'a, 'tree>(
         &self,
         tree: &'tree Tree,
+        source_path: &'tree Path,
         source: &'tree str,
+        tsg_path: &'a Path,
+        tsg_source: &'a str,
```
hendrikvanantwerpen, line 71: It _is_ possible to construct the `File` data structure directly and not from a source. However, one would also have to fake all the `location`s in there. Anyway, perhaps it is worth considering if anything panics if an empty source is passed here. (Panicking in error reporting seems something worth avoiding.)

src/execution.rs
hendrikvanantwerpen, line 419: This could be replaced by a call to this helper function: https://github.com/tree-sitter/tree-sitter-graph/blob/main/src/execution.rs#L955-L986.

src/execution.rs
hendrikvanantwerpen: I don't know either.

src/lazy_execution.rs
```diff
@@ -162,9 +174,28 @@ impl ast::Stanza {
...
-                .execute_lazy(&mut exec)
-                .with_context(|| format!("Executing {}", statement))?;
+            statement.execute_lazy(&mut exec).or_else(|e| {
+                Err(e).with_context(|| {
+                    let node = mat
+                        .captures
+                        .iter()
+                        .find(|c| c.index as usize == self.full_match_file_capture_index)
+                        .unwrap()
+                        .node;
```
hendrikvanantwerpen, line 184: Same helper function as above: https://github.com/tree-sitter/tree-sitter-graph/blob/main/src/execution.rs#L955-L986

hendrikvanantwerpen (commented):
This is very nice and I think the idea of having a systematic format statement-in_stanze-matching_x is great! That seems like a good amount of context for a CLI.

I do wonder if the decision of how to display errors is in the wrong place though, for the following reasons:

1. The construction of error messages is part of the execution code. Especially since the formatting is now more involved, that seems like a separate concern to me. 
2. The API now assumes that the TSG and parse tree have file-with-a-path origins. This is not always the case. In Aleph the source comes from a remote call and the TSG source is embedded in the binary, not read from a file. I think the API is cleaner if we don't force users to always come up with values for those.
3. The optimal format of reported errors depends a lot on the context. In the CLI, having multi-line colored output with inline fragments is fantastic. In Aleph, with errors ending up in Splunk, a more compact one-line error without coloring might be much more useful.

In light of this, let me suggest a slightly different setup. Errors carry two locations, like your rejected suggestion below, but the execution code does not do any formatting. Instead, the formatting is left to the user of the API, and we can provide different methods, for example `.fmt_compact` and `.fmt_verbose` (just the first to come to mind), where `.fmt_verbose` produces the multi-line errors, and takes the sources and paths as parameters that it needs to produce the error. The `.fmt_compact` perhaps takes no parameters (and could also be the `Display` implementation).

How does this sound?

src/execution.rs
robrix: Then at least I'm in good company!

src/execution.rs
robrix: AFAICS no. There (I believe) only three new `unwrap`s, and the only unsafe one occurs when retrieving the root node for a match; if it panics, it's because one of our or `tree-sitter`'s invariants has failed.

The other two are when formatting `Excerpt`s, whose `source` is `Option`al (because what if there is no _n_th line of the indicated string?), and whose `path` is non-`Option`al but must be converted back to a `str` for display (which can fail). In both of these cases we unwrap with a default (`unwrap_or`).

src/execution.rs
robrix: Thank you!

robrix:
> I do wonder if the decision of how to display errors is in the wrong place though, for the following reasons:
> 
> 1. The construction of error messages is part of the execution code. Especially since the formatting is now more involved, that seems like a separate concern to me.

It is, absolutely, tho I'd note that this was the case before this PR: every one of `ExecutionError`'s constructors except `Other` wraps a `String`.

> 2. The API now assumes that the TSG and parse tree have file-with-a-path origins. This is not always the case. In Aleph the source comes from a remote call and the TSG source is embedded in the binary, not read from a file. I think the API is cleaner if we don't force users to always come up with values for those.

Excellent point.

> 3. The optimal format of reported errors depends a lot on the context. In the CLI, having multi-line colored output with inline fragments is fantastic. In Aleph, with errors ending up in Splunk, a more compact one-line error without coloring might be much more useful.

Likewise, excellent point. I'm glad to know this.

> In light of this, let me suggest a slightly different setup. Errors carry two locations, like your rejected suggestion below, but the execution code does not do any formatting. Instead, the formatting is left to the user of the API, and we can provide different methods, for example `.fmt_compact` and `.fmt_verbose` (just the first to come to mind), where `.fmt_verbose` produces the multi-line errors, and takes the sources and paths as parameters that it needs to produce the error. The `.fmt_compact` perhaps takes no parameters (and could also be the `Display` implementation).
> 
> How does this sound?

I like this plan, tho I will note it's fairly invasive. Let's see how it goes.

hendrikvanantwerpen:
> It is, absolutely, tho I'd note that this was the case before this PR: every one of ExecutionError's constructors except Other wraps a String.

Fair point 🙂.

> I like this plan, tho I will note it's fairly invasive. Let's see how it goes.

I understand and if this is distracting you from more urgent things we can consider merging this as is and doing that in a separate PR. Then we can discuss who makes the most sense to do that work.

robrix:
> > It is, absolutely, tho I'd note that this was the case before this PR: every one of ExecutionError's constructors except Other wraps a String.
> 
> Fair point 🙂.

I think this is made somewhat inevitable by the design of `anyhow::Error`, which I think hints that it's not a great fit for what we're doing here.

> > I like this plan, tho I will note it's fairly invasive. Let's see how it goes.
> 
> I understand and if this is distracting you from more urgent things we can consider merging this as is and doing that in a separate PR. Then we can discuss who makes the most sense to do that work.

After today's exploration, I think that's our best bet, tho I'm also not in any special hurry since I can use this locally just fine.

hendrikvanantwerpen:
> I think this is made somewhat inevitable by the design of `anyhow::Error`, which I think hints that it's not a great fit for what we're doing here.

That is a very good point. As a general rule we don't use `anyhow` errors in library code, but I used it here as a quick way to get the context mechanism working. I guess we'd have to reimplement that as part of our error type if we want to have more control over our error output...

hendrikvanantwerpen (approved):
I am going ahead and merge this. I'll start work on restructuring the error handling to get rid of `anyhow::Error` in the library code in a follow up PR.

PR #94, opened 2022-08-18
Add cancellation

hendrikvanantwerpen:
Add cancellation support.

This changes the API: `ast::File::execute` now has an extra parameter for the cancellation flag.

_Reviewers are encouraged to take a look at the other PRs listed in https://github.com/github/aleph/issues/1641 as well, to see how this design plays out through the whole stack._

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -275,15 +275,16 @@ pub trait CancellationFlag {
     fn check(&self, at: &'static str) -> Result<(), CancellationError>;
 }
 
-pub struct NoCancellation;
-impl CancellationFlag for NoCancellation {
-    fn check(&self, _at: &'static str) -> Result<(), CancellationError> {
-        Ok(())
-    }
-}
-impl Default for NoCancellation {
-    fn default() -> Self {
-        Self
+pub struct CancellationFlags;
+impl CancellationFlags {
+    pub fn none() -> impl CancellationFlag {
+        struct NoCancellation;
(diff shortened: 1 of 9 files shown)


src/execution.rs
```diff
@@ -263,6 +270,29 @@ impl ExecutionError {
...
+    pub fn none() -> impl CancellationFlag {
+        struct NoCancellation;
+        impl CancellationFlag for NoCancellation {
+            fn check(&self, _at: &'static str) -> Result<(), CancellationError> {
+                Ok(())
+            }
+        }
+        NoCancellation
+    }
+}
```
dcreager, line 289: You could expose `NoCancellation` directly instead of having the factory function.  That would change callers from `&CancellationFlags::none()` to `&NoCancellation`.  I have a slight aesthetic preference for exposing the type directly, unless there's something that I'm missing.

dcreager (approved):
This is great! :+1: especially to using a `&dyn` reference instead of static type parameters.

src/execution.rs
hendrikvanantwerpen: In an earlier iteration `NoCancellation` had parameters and this was a neat way to hide all that from the user. I would have preferably written `none()` as an associated function of `CancellationFlag` but that is not technically possible.

I don't think there is much of a difference between the two options now, so I'm fine either way.

dcreager (approved):

PR #98, opened 2022-09-01
Improve error handling

hendrikvanantwerpen:
The library code used `anyhow::Error` to implement adding context toerrors. This resulted in the loss of cancellation errors, which are an important signal for some use cases.

This PR eliminates `anyhow` completely from the library code and implements a custom error case to register context. We ensure that cancellation errors are never wrapped in context, but propagated as is.

This PR also factors out the fancy formatting logic into a dedicated `.display_pretty()` method, which gives users a choice how to format errors.


src/bin/tree-sitter-graph/main.rs
```diff
@@ -138,17 +138,13 @@ fn main() -> Result<()> {
...
-            tsg_path,
-            &tsg,
-            &mut config,
-            &NoCancellation,
-        )
-        .with_context(|| format!("Cannot execute TSG file {}", tsg_path.display()))?;
+    let graph = match file.execute(&tree, &source, &mut config, &NoCancellation) {
+        Ok(graph) => graph,
+        Err(e) => {
+            eprint!("{}", e.display_pretty(source_path, &source, tsg_path, &tsg));
```
dcreager, line 144: Generally it's a code smell to `fprintf` from a library like this.  Is this message intended to eventually make it into Splunk?  Do we need to bite the bullet and pull in the `log` or `tracing` crate so that callers can control whether and how these messages are printed?  (And if so, I'm fine with that being a TODO comment here + a follow-on PR)

(If we do go down that route, `tracing` would be my preference, because then we could configure the log output to match what's happening on the Go side, and we'd end up with nicely structured log messages from both Go and Rust in the same place)

src/execution_error.rs
```diff
@@ -0,0 +1,275 @@
...
+                }?;
+                self.fmt_entry(f, index + 1, cause)
+            }
+            other => writeln!(f, "{:>5}: {}", index, other),
+        }
+    }
+}
+
+/// Excerpts of source from either the target language file or the tsg rules file.
+pub struct Excerpt<'a> {
```
dcreager, line 210: I realize this isn't introducing this code, just moving it, but we might consider using the [`miette`](https://docs.rs/miette/*/miette/) crate for constructing "nested context based on source code".  It has a very nice [data model](https://docs.rs/miette/5.3.0/miette/#-snippets) for these kinds of errors, and also has a good ability to separate error construction from error display/reporting.

src/bin/tree-sitter-graph/main.rs
hendrikvanantwerpen: This is the CLI code, not the library, so I think it should be okay here. For library code, I completely agree, though.

dcreager: Oh yes you're right!  I withdraw my objection :smile:

src/execution_error.rs
hendrikvanantwerpen: I have a déjà-vu. I feel like this was discussed before, and my reply is the same as last time 😛: If we want this, let's do it in another PR.

dcreager (approved):

src/execution_error.rs
dcreager: Agreed

PR #99, opened 2022-09-01
Reorganize execution code

hendrikvanantwerpen:
The modules for the execution code were becoming a little messy.

This PR reorganizes that code to have a single top-level execution module
(instead of `execution` and `lazy_execution`), and have submodules for
`strict`, `lazy`, and `error` logic.


src/execution.rs
```diff
@@ -93,48 +57,8 @@ impl File {
...
-                &mut cursor,
-                &self.shorthands,
-                cancellation_flag,
-            )?;
-        }
-        Ok(())
-    }
 
-    pub(crate) fn check_globals(&self, globals: &Globals) -> Result<(), ExecutionError> {
+    pub(self) fn check_globals(&self, globals: &Globals) -> Result<(), ExecutionError> {
```
dcreager, line 61: TIL `pub(self)`!  Is that "this module and nested modules, but not any other modules in the crate"?

dcreager (approved):

src/execution.rs
hendrikvanantwerpen: Indeed! The visibility in Rust is quite nice like that. It gives you the possibility to split things in different modules without having to expose it to unrelated modules in the same crate.

PR #103, opened 2022-09-19
Terminal coloring dependencies

hendrikvanantwerpen:
The `ansi_term` library used for terminal coloring is unmaintained: https://github.com/advisories/GHSA-74w3-p89x-ffgh.

This PR:
- Replaces it with the `colored` library.
- Puts terminal coloring behind the `term-colors` feature flag to reduce the number of required dependencies for library users.
- Adds a CI step that builds the project with all features enabled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,8 +33,10 @@ jobs:
           key: ${{ runner.OS }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: |
             ${{ runner.OS }}-cargo-
-      - name: Build library
+      - name: Build library (default features)
         run: cargo build
+      - name: Build library (all features)
+        run: cargo build --all-features
       - name: Build program
         run: cargo build --bin tree-sitter-graph --features=cli
       - name: Run test suite
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,9 +33,13 @@ jobs:
           key: ${{ runner.OS }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: |
             ${{ runner.OS }}-cargo-
-      - name: Build & test library (default features)
+      - name: Build library (default features)
+        run: cargo build
+      - name: Test library (default features)
         run: cargo test
-      - name: Build & test library (all features)
+      - name: Build library (all features)
+        run: cargo build --all-features
+      - name: Test library (all features)
         run: cargo test --all-features
       - name: Build program
         run: cargo build --bin tree-sitter-graph --features=cli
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -33,8 +33,10 @@ jobs:
           key: ${{ runner.OS }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: |
             ${{ runner.OS }}-cargo-
-      - name: Build library
+      - name: Build library (default features)
         run: cargo build
+      - name: Build library (all features)
+        run: cargo build --all-features
```
hendrikvanantwerpen, line 39: I wonder what a sensible CI strategy is here.

First about which feature combinations to test. Adding an all features build is I think a good idea, because it future-proofs us against forgetting to add new feature flags to CI. Testing individual featurescould also be useful, though, to ensure the features are independent or specify the correct dependencies.

Second about having build and test targets. If I understand the [docs](https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection) correctly, test implies build. It might be enough to only have test commands (with the desired feature combinations).

hendrikvanantwerpen: My proposal in 4438ad732bf7ec2ec541bc14116e36714bcae6bc: run `cargo test` with default and all features, and build the CLI individually as we did before.

dcreager: > Second about having build and test targets

I had these separate mostly so that we would see compilation failures from the library code itself earlier / in a separate section.  You're right that `cargo test` will build whatever other stuff is necessary, so if we're finding that separation not to be useful, I'm :+1: to how you've reduced the number of steps.

(Though note that keeping them separate should not increase the CI time, since the build results from the separate `cargo build` step will be reused during the `cargo test` step.)

Cargo.toml
```diff
@@ -16,7 +16,6 @@ edition = "2018"
 test = false
 
 [dependencies]
-ansi_term = "0.12"
```
dcreager, line 19: :+1: this is a very good cleanup

src/execution/error.rs
```diff
@@ -266,10 +261,39 @@ impl<'a> std::fmt::Display for Excerpt<'a> {
...
 }
+
+// coloring functions
+
+#[cfg(feature = "term-colors")]
+fn blue(str: &str) -> impl std::fmt::Display {
+    str.blue()
+}
+#[cfg(not(feature = "term-colors"))]
+fn blue<'a>(str: &'a str) -> impl std::fmt::Display + 'a {
```
dcreager, line 279: nit: Do you need the explicit lifetime bounds here?  Does it work to give this the same signature as the `term-colors` variant above?

dcreager (approved):

.github/workflows/ci.yml
hendrikvanantwerpen: That's a good point, about the caching. I think I'll split them up after all then.

PR #110, opened 2022-09-30
Add support for default string values for global variables

### LINKED ISSUE #104: Default values for global variables
Global variables sometimes have sensible default values. It should the possible to provide those in the TSG.

I suggest the following way of doing that:

    global VAR = VALUE

The `= VALUE` part is optional, and globls without it must always be provided. The value is only used if the global variable is not set by the caller.

hendrikvanantwerpen:
Adds support for default string values for global variables. For example:

    global PKG_NAME = ""

Do we want to allow more types of values? A more flexible design (but more invasive change) would be to allow arbitrary expressions, so that you can write something like:

    global FILE_NAME = ""
    global MODULE_NAME = (replace "/" "." FILE_NAME)

Fixes #104.

## Prerequisites

- [ ] Depends on #109.

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -58,13 +58,24 @@ impl File {
         }
     }
 
-    pub(self) fn check_globals(&self, globals: &Globals) -> Result<(), ExecutionError> {
+    pub(self) fn check_globals(&self, globals: &mut Globals) -> Result<(), ExecutionError> {
         for global in &self.globals {
             match globals.get(&global.name) {
                 None => {
-                    return Err(ExecutionError::MissingGlobalVariable(
-                        global.name.as_str().to_string(),
-                    ));
+                    if let Some(default) = &global.default {
+                        globals
+                            .add(global.name.clone(), default.to_string().into())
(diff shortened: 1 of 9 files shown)


src/execution.rs
```diff
@@ -58,13 +58,24 @@ impl File {
...
-                    ));
+                    if let Some(default) = &global.default {
+                        globals
+                            .add(global.name.clone(), default.to_string().into())
+                            .map_err(|_| {
+                                ExecutionError::DuplicateVariable(format!(
+                                    "global variable {} already defined",
+                                    global.name
+                                ))
+                            })?;
```
dcreager, line 73: This could maybe be a panic, or an `unreachable!()` inside the `map_err` callback, since this code only gets executed when `globals.get` fails.  (Though on the other hand, this is safer, and I can't imagine the redundant check adds enough overhead to insist on removing it.)

dcreager (approved):
> Do we want to allow more types of values?

I like how your proposed extended syntax is a superset of what's here, so that we can punt on this for now.  :smile:

PR #112, opened 2022-10-18
Add some useful standard library functions

hendrikvanantwerpen:
This adds a couple of useful functions to the standard library before the next release.

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -131,6 +131,7 @@ impl Functions {
         functions.add(Identifier::from("format"), stdlib::string::Format);
         functions.add(Identifier::from("replace"), stdlib::string::Replace);
         // list functions
+        functions.add(Identifier::from("concat"), stdlib::list::Concat);
         functions.add(Identifier::from("is-empty"), stdlib::list::IsEmpty);
         functions.add(Identifier::from("length"), stdlib::list::Length);
         functions
@@ -580,6 +581,24 @@ pub mod stdlib {
 
     pub mod list {
         use super::*;
+
+        /// The implementation of the standard [`concat`][`crate::reference::functions#concat`] function.
+        pub struct Concat;
+
+        impl Function for Concat {
+            fn call(
+                &self,
+                _graph: &mut Graph,
+                _source: &str,
+                parameters: &mut dyn Parameters,
+            ) -> Result<Value, ExecutionError> {
+                let mut result = Vec::new();
+                while let Ok(list) = parameters.param() {
+                    result.append(&mut list.into_list()?);
+                }
+                Ok(result.into())
+            }
+        }
 
         /// The implementation of the standard [`is-empty`][`crate::reference::functions#is-empty`] function.
         pub struct IsEmpty;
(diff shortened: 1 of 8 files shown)


src/functions.rs
```diff
@@ -169,6 +173,71 @@ pub mod stdlib {
...
+                    _ => {}
+                },
+            };
+            Err(ExecutionError::FunctionFailed(
+                "eq".into(),
+                format!(
+                    "Cannot compare values of different types: {} and {}",
+                    left, right
+                ),
+            ))
```
dcreager, line 237: I was about to ask if you could use the derived `std::ops::Eq` impl for this, but now I can see why you didn't!

src/reference/functions.rs
```diff
@@ -14,12 +14,19 @@
 //!
 //! # General functions
 //!
+//! ## `eq`
+//!
+//! Check if values are equal.
```
dcreager, line 19: This deserves a sentence about how you're handling nulls.

src/functions.rs
```diff
@@ -450,6 +519,43 @@ pub mod stdlib {
...
+                            None => return Err(ExecutionError::FunctionFailed("format".into(), format!("Unexpected end of format string `{}` after `{{`. Expected `{{` or `}}`.", format))),
+                        },
+                        '}' => match it.next() {
+                            Some((_, '}')) => result.push('}'),
+                            Some((i, c)) => return Err(ExecutionError::FunctionFailed("format".into(), format!("Unexpected character `{}` after `}}` at position {} in format string `{}`. Expected `}}`.", c, i + 1, format))),
+                            None => return Err(ExecutionError::FunctionFailed("format".into(), format!("Unexpected end of format string `{}` after `{{. Expected `}}`.", format))),
+                        },
+                        c => result.push(c),
+                    }
+                }
```
dcreager, line 553: If we ever want to support modifiers in the format strings, there's a [`runtime_fmt`](https://docs.rs/runtime-fmt/latest/runtime_fmt/) crate that might be useful.  No need to bring that in yet, though, this is :+1: for the simplified format strings that you're currently supporting.

src/functions.rs
```diff
@@ -477,6 +583,24 @@ pub mod stdlib {
...
+        impl Function for Concat {
+            fn call(
+                &self,
+                _graph: &mut Graph,
+                _source: &str,
+                parameters: &mut dyn Parameters,
+            ) -> Result<Value, ExecutionError> {
+                let mut result = Vec::new();
+                while let Ok(list) = parameters.param() {
+                    result.append(&mut list.into_list()?);
```
dcreager, line 598: Very nice, I've always done this with `extend_from_slice`, didn't realize there was an `append` method!

dcreager (approved):

src/functions.rs
hendrikvanantwerpen: I did at first, and then decided that I didn't want to be responsible for that kind of language :P

src/functions.rs
hendrikvanantwerpen: Thanks for that reference! I was looking for something like that and couldn't find it. It was indeed the purpose to be open to richer specifiers in the future.

PR #114, opened 2022-10-18
Add job to create GitHub release

hendrikvanantwerpen:
This adds an workflow to create a GitHub release whenever a new version tag is pushed.

Action docs: https://github.com/marketplace/actions/create-release

Depends on #113.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,16 +1,16 @@
-name: Publish on crates.io
+name: Publish release
 
 on:
   push:
     tags:
-      - v*'
+      - v*
 
 env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
 
 jobs:
-  publish:
+  publish-crate:
     runs-on: ubuntu-latest
     steps:
       - name: Install Rust environment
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -26,3 +25,14 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Create GitHub release
+        uses: ncipollo/release-action@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
rewinfrey, line 38: I see the `create-release` action offers the option to also upload an artifact,  and it makes me wonder if that would useful to provide a tar binary per release?

rewinfrey (approved):
I think this looks good as is,  but left a question for you to consider about including a release artifact.

.github/workflows/publish.yml
hendrikvanantwerpen: I do like this idea. I'm not entirely sure how to implement ort though. We'd want to provide artifacts for multiple platforms, I presume. They would be different jobs (in a matrix?), that all contribute an artifact to the release. I don't know if that would work with this plugin. Perhaps I'll do an experiment tomorrow to see if the runs can be made to combine, or they will conflict.

dcreager: I believe the we're doing release artefacts in tree-sitter so that the `npm` installer has something easy to download and install.  But this crate is largely installed via `cargo install`, and so I think it's fine to create the release without uploading anything to it.  Put another way, publishing to crates.io is the equivalent to / replacement for attaching a tarball to the GitHub release.

dcreager (approved):

.github/workflows/publish.yml
hendrikvanantwerpen: In that case, let's not put the effort for now.

PR #121, opened 2023-03-10
Fixes for error reporting

hendrikvanantwerpen:
This addresses two issues with error reporting:

1. Some errors during strict evaluation cause panics during error formatting. The problem was a use of the wrong index.

2. Several `unwrap` calls are replaced by `expect` to make debugging easier when similar problems are reported.


robrix (approved):
Very yes 👍🏻

PR #122, opened 2023-03-29
Improve sharing of functions

hendrikvanantwerpen:
Trait instances stored in `Functions` are difficult to share in concurrent situations. Apparently the `'static` lifetime is not enough for that. This PR adds `Send + Sync` bounds to improve that.


patrickt (approved):
Weird. But I don’t have enough Rust expertise to disagree, so… LGTM!

PR #124, opened 2023-03-31
Pretty errors for lazy execution

hendrikvanantwerpen:
Strict execution has had pretty errors with source excerpts for some time now. This PR makes the changes to support the same for lazy execution. This is useful because the tree-sitter-stack-graphs tool uses lazy execution.

### DIFF
--- a/src/execution/error.rs
+++ b/src/execution/error.rs
@@ -266,7 +266,7 @@ impl<'a> Excerpt<'a> {
 impl<'a> std::fmt::Display for Excerpt<'a> {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
         // path and line/col
-        writeln!(
+        write!(
             f,
             "{}{}:{}:{}:",
             " ".repeat(self.indent),
@@ -275,6 +275,7 @@ impl<'a> std::fmt::Display for Excerpt<'a> {
             white_bold(&format!("{}", self.location.column + 1)),
         )?;
         if let Some(source) = self.source {
+            writeln!(f)?;
             // first line: line number & source
             writeln!(
                 f,
@@ -295,7 +296,7 @@ impl<'a> std::fmt::Display for Excerpt<'a> {
                 green_bold("^")
             )?;
         } else {
-            writeln!(f, "{}{}", " ".repeat(self.indent), "<missing source>",)?;
+            writeln!(f, " <missing source>")?;
         }
         Ok(())
     }
--- a/src/execution/error.rs
+++ b/src/execution/error.rs
@@ -151,7 +151,10 @@ impl<R> ResultWithExecutionError<R> for Result<R, ExecutionError> {
     {
         self.map_err(|e| match e {
             cancelled @ ExecutionError::Cancelled(_) => cancelled,
-            in_context @ ExecutionError::InContext(_, _) => in_context,
+            in_other_context @ ExecutionError::InContext(Context::Other(_), _) => {
+                ExecutionError::InContext(with_context(), Box::new(in_other_context))
+            }
+            in_stmt_context @ ExecutionError::InContext(_, _) => in_stmt_context,
             _ => ExecutionError::InContext(with_context(), Box::new(e)),
         })
     }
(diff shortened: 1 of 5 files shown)


src/execution/lazy.rs
```diff
@@ -354,24 +348,20 @@ impl ast::Scan {
...
                 shorthands: exec.shorthands,
                 cancellation_flag: exec.cancellation_flag,
             };
 
             for statement in &arm.statements {
-                arm_exec
-                    .cancellation_flag
-                    .check("executing scan statements")?;
+                arm_exec.error_context.statement = format!("{}", statement);
+                arm_exec.error_context.statement_location = statement.location();
```
hendrikvanantwerpen, line 358: The `StatementContext` is updated in nested blocks, so that it always represents the innermost statement.

src/execution/lazy/statements.rs
```diff
@@ -35,10 +36,18 @@ impl LazyStatement {
         debug!("eval {}", self);
         trace!("{{");
         let result = match self {
-            Self::AddGraphNodeAttribute(stmt) => stmt.evaluate(exec),
-            Self::CreateEdge(stmt) => stmt.evaluate(exec),
-            Self::AddEdgeAttribute(stmt) => stmt.evaluate(exec),
-            Self::Print(stmt) => stmt.evaluate(exec),
+            Self::AddGraphNodeAttribute(stmt) => stmt
+                .evaluate(exec)
+                .with_context(|| stmt.debug_info.clone().into()),
```
hendrikvanantwerpen, line 41: During the second phase, where the lazy values are evaluated, the stored context is attached to any errors that are thrown. Here when a statement is executed.

src/execution/lazy/store.rs
```diff
@@ -78,10 +78,8 @@ impl LazyStore {
         exec: &mut EvaluationContext,
     ) -> Result<graph::Value, ExecutionError> {
         let variable = &self.elements[variable.store_location];
-        let debug_info = variable.debug_info;
-        let value = variable.force(exec).with_context(|| {
-            format!("via {} at {}", (*variable.state.borrow()), debug_info).into()
-        })?;
+        let debug_info = variable.debug_info.clone();
+        let value = variable.force(exec).with_context(|| debug_info.0.into())?;
```
hendrikvanantwerpen, line 82: Here when a value is loaded from the store. Since values originate from other statements (the ones where they were bound), this ensures that an evaluation error reports on the statement that constructed the value, not the statement that forced its evaluation.

src/execution/lazy/store.rs
```diff
@@ -148,13 +146,10 @@ impl LazyScopedVariables {
...
-                        format!(
-                            "Evaluating scope of variable _.{} set at {}",
-                            name, debug_info
-                        )
-                        .into()
-                    })?;
+                    let node = scope
+                        .evaluate_as_syntax_node(exec)
+                        .with_context(|| format!("Evaluating scope of variable _.{}", name,).into())
+                        .with_context(|| debug_info.0.clone().into())?;
```
hendrikvanantwerpen, line 152: Similarly for scoped variables.

src/execution/error.rs
```diff
@@ -124,6 +151,10 @@ impl<R> ResultWithExecutionError<R> for Result<R, ExecutionError> {
     {
         self.map_err(|e| match e {
             cancelled @ ExecutionError::Cancelled(_) => cancelled,
+            in_other_context @ ExecutionError::InContext(Context::Other(_), _) => {
+                ExecutionError::InContext(with_context(), Box::new(in_other_context))
+            }
+            in_stmt_context @ ExecutionError::InContext(_, _) => in_stmt_context,
```
hendrikvanantwerpen, line 157: We only want to report the innermost statement (in phase one), or the originating statement (in phase two). Therefore we ignore any statement context if another statement context is already attached to the error. Other contexts (which are strings) do not prevent attaching a statement context.

hendrikvanantwerpen (commented):
I have sketched the implementation idea in comments.

_This is a little experiment. I think comments appear in the conversation view in the order they were created. In that case, they should be in a sensible order. In the diff view they'll be in a weird order, though._

patrickt (approved):
LGTM. Love some pretty errors.

PR #133, opened 2023-05-15
Expose method to execute TSG matches without executing stanzas

hendrikvanantwerpen:
This PR exposes a method to match stanza queries without also executing stanza bodies. This has essentially the same effect as running `tree-sitter query` on a query file containing all stanza queries. This will be used in github/stack-graphs to implement a command to emulate `tree-sitter query`.

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -118,9 +118,10 @@ impl File {
                         let index = file_query
                             .capture_index_for_name(name)
                             .expect("missing index for capture");
-                        (name, index)
+                        let quantifier = file_query.capture_quantifiers(0)[index as usize];
+                        (name, quantifier, index)
                     })
-                    .filter(|c| c.1 != stanza.full_match_file_capture_index as u32)
+                    .filter(|c| c.2 != stanza.full_match_file_capture_index as u32)
                     .collect();
                 visit(Match {
                     mat,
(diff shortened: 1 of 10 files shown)


src/execution.rs
```diff
@@ -93,6 +96,154 @@ impl File {
...
+        &self,
+        tree: &'tree Tree,
+        source: &'tree str,
+        lazy: bool,
+        mut visit: F,
+    ) -> Result<(), E>
+    where
+        F: FnMut(Match<'_, 'tree>) -> Result<(), E>,
+    {
+        if lazy {
```
dcreager, line 110: As an aside, do we need to continue having both strict and lazy evaluation modes?  IIRR, we kept them both as you were first implementing lazy evaluation to give us a gradual migration strategy.  But now that lazy is working in production, do we need to keep strict around?

dcreager (approved):
Looks good, just one unrelated question

src/execution.rs
hendrikvanantwerpen: Good question. If I recall, @tausbn was using some mutable state tricks that only work in strict more. Maybe he can weigh in on that?

tausbn: By my count, we have _two_ remaining mutable state tricks left. (See backlink to the relevant issue below.)

I think it should be possible to get rid of them, though that has not been a priority so far.

PR #135, opened 2023-06-01
Inherited scoped variables

### LINKED ISSUE #88: "Scoped variable" is a misnomer
In the old DSL, scoped variables were available within the lexical scope of the node they were attached to—its entire subtree.

This is no longer the case, leading me to believe that the name is a misnomer, and that we should try to come up with something more indicative of what they mean.

hendrikvanantwerpen:
This PR adds support for inherited scoped variables. Until now, scoped variables were attached to a single node. With this PR it is possible to add `inherit .var_name` to a file, which will cause `@node.var_name` to look for the scoped variable or on the node or any of its parents (selecting the closest parent that has a value for that variable).

This adds an easy mechanism to propagate information from parent to child nodes without having to write a lot of verbose code to do it.

Fixes #88.


rewinfrey (approved):
👍, this will be very nice to use, thank you!

PR #138, opened 2023-07-12
Lazy node creation

hendrikvanantwerpen:
This PR changes node creation in lazy execution mode to be lazy instead of eager. Previously, all nodes were created and added to the graph, regardless of whether they were used.

Having the option to exclude unused nodes from the graph can make it easier to defined commonly used nodes on all syntax nodes, with something like this:

    (_)@node {
        node @node.one_node
        node @node.another_node
        ...
    }


hendrikvanantwerpen:
This was born out of an attempt at getting something out quick, but might perhaps be facilitating an anit-pattern. Closing it for now.

PR #140, opened 2023-07-17
Add `match_node_attr`

BekaValentine:
This PR adds a new match node attribute to support better debugging. The node being stored is the syntax node that was matched on to give rise to the graph node in question.


hendrikvanantwerpen (approved):
LGTM

PR #149, opened 2024-03-08
Allow duplicate edges and attributes

### LINKED ISSUE #85: Allow duplicate edges
Propagators have a neat design whereby each node's value is either unset, set to a specific value, or conflicting. You can set a value by propagating something to it; if two values are propagated to it, and they are equal, then it remains unchanged with that value; and if the values differ, it's in conflict—an error.

I'd like to see something similar for `tree-sitter-graph` edges: setting the same edge is harmless, and thus accepted. Setting the same attr with the same value on that edge is also harmless, and thus accepted. Distinct values for the attr are an error.

(I'd actually like to see something like this for nodes and their attrs as well, but nodes have identity, which makes things more annoying.)

### LINKED ISSUE #93: Allow duplicate attrs
Like #85, but for attributes: using something like propagators, allow redundant attributes &c. if and only if their values are compatible (for most purposes we can think of this as "equal-valued").

hendrikvanantwerpen:
Relaxes runtime checks so that setting an edge multiple times, or setting attributes to the same value multiple times, is no longer an error.

Fixes #85 and #93.

If we allow this for edges and attributes, it makes sense to also allow this for `node @capture.name` occurring multiple times for the same syntax node. I've not addressed that here because implementing that is more involved and requires tracking some extra bookkeeping during execution.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tree-sitter-graph"
-version = "0.11.1"
+version = "0.11.2"
 description = "Construct graphs from parsed source code"
 homepage = "https://github.com/tree-sitter/tree-sitter-graph/"
 repository = "https://github.com/tree-sitter/tree-sitter-graph/"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "tree-sitter-graph"
-version = "0.11.1"
+version = "0.11.2"
```
dcreager, line 3: It might be a borderline case, but I consider this enough to cause a minor bump to 0.12

dcreager (approved):

Cargo.toml
dcreager: (per slack, I only feel mildly strongly about this, fine with a patch bump too)

PR #156, opened 2024-06-12
Allow optional globals be omitted

bm-w:
When defining an optional global, currently the library still requires that it be set in the [`ExecutionConfig::globals`][] passed to [`File::execute`][], or it fails with an [`ExecutionError(MissingGlobalVariable("MY_GLOBAL"))`][err]. This change allows the global to be omitted:

```tsg
constant MY_GLOBAL?
```

```rust
let globals = Variables::new(); // Empty; `MY_GLOBAL` does not need to be set
let config = ExecutionConfig::new(&functions, &globals);
let graph = file.execute(tree, source, &config, &NoCancellation)?;
```

[`ExecutionConfig::globals`]: https://github.com/tree-sitter/tree-sitter-graph/blob/v0.11.2/src/execution.rs#L255
[`File::execute`]: https://github.com/tree-sitter/tree-sitter-graph/blob/v0.11.2/src/execution.rs#L35
[err]: https://github.com/tree-sitter/tree-sitter-graph/blob/v0.11.2/src/execution/error.rs#L51

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -68,9 +68,17 @@ impl File {
         for global in &self.globals {
             match globals.get(&global.name) {
                 None => {
-                    if let Some(default) = &global.default {
+                    let default_or_null = global
+                        .default
+                        .as_ref()
+                        .map(|default| default.to_string().into())
+                        .or_else(|| {
+                            (global.quantifier == CaptureQuantifier::ZeroOrOne)
+                                .then_some(Value::Null)
+                        });
+                    if let Some(default_or_null) = default_or_null {
                         globals
-                            .add(global.name.clone(), default.to_string().into())
+                            .add(global.name.clone(), default_or_null)
                             .map_err(|_| {
                                 ExecutionError::DuplicateVariable(format!(
                                     "global variable {} already defined",
--- a/tests/it/execution.rs
+++ b/tests/it/execution.rs
@@ -291,6 +291,28 @@ fn can_omit_global_variable_with_default() {
 }
 
 #[test]
+fn can_omit_optional_global_variable() {
+    check_execution(
+        "pass",
+        indoc! {r#"
+          global pkgname?
+
+          (module)
+          {
+            node n
+            if (is-null pkgname) {
+              attr (n) pkgname = "fallback"
+            }
+          }
+        "#},
+        indoc! {r#"
+          node 0
+            pkgname: "fallback"
+    "#},
+    );
+}
+
+#[test]
 fn cannot_omit_global_variable() {
     fail_execution(
         "pass",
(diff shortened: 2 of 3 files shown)


hendrikvanantwerpen (commented):
I don't think there's a need for custom logic here. Setting a default value will di the trick here:

    global foo? = #null

I checked and see that this feature is missing from the reference documentation. I'll look into that.

bm-w:
Actually, that gives me a parsing error:
```tsg
global foo? = #null
```
```plain
Error: parsing TSG file from "test.tsg"

Caused by:
    Expected '"' at (1, 15)
```
Seems globals must be strings?

bm-w:
I was looking into the global default parsing, and that could be changed to supported literals more along the lines of:
https://github.com/tree-sitter/tree-sitter-graph/blob/68504d72049b131064941fff105e5deaf8ea7074/src/parser.rs#L326
```diff
--default = Some(self.parse_string()?);
++if self.peek()? == '#' {
++    default = Some(match self.parse_literal()? {
++        ast::Expression::TrueLiteral => Value::Boolean(true),
++        ast::Expression::NullLiteral => {
++            if quantifier == CaptureQuantifier::ZeroOrOne {
++                Value::Null
++            } else {
++                todo!("handle unexpected null")
++            }
++        }
++        _ => unreachable!(),
++    });
++} else {
++    default = Some(self.parse_string()?.into());
++}
```
Where I also changed `ast::Global::default` from `Option<String>` to `Option<Value>`:
```diff
/// A global variable
#[derive(Debug, Eq, PartialEq)]
  pub struct Global {
      /// The name of the global variable
      pub name: Identifier,
      /// The quantifier of the global variable
      pub quantifier: CaptureQuantifier,
      /// Default value
--    pub default: Option<String>,
++    pub default: Option<Value>,
      pub location: Location,
  }
```

I’d be happy to close this PR and open that one? Although I kind of feel it’d be nice to allow users to omit the `#null` (i.e. I’d say merge this PR and also make this change to allow literals as global defaults).

bm-w:
Hi, @hendrikvanantwerpen! I just noticed this PR still open — did you manage to make any progress on the idea you last said you had?

hendrikvanantwerpen:
@bm-w Thanks for reminding me. I didn't (and don't) have time to do the patch I had in might. I think your proposal in https://github.com/tree-sitter/tree-sitter-graph/pull/156#issuecomment-2163035320 is good.

I would suggest to keep it simple:
- Don't try to infer defaults (like `#null` for an optional), always require it to be set explicitly.
- Accept any value, so if the user sets `#null` use it. Using the suffixes (like `?` or `*`) to check if `#null` is valid is quite weak anyway, so it's not going to be 100% safe whatever we do.

